PR #1, opened 2021-07-30
Add Context diagram

ccremer:


tobru (changes requested):
Why do you remove the navigation structure? I'd like to keep that structure to have it the same on all documentation sites. And it would break the integration into kb.vshn.ch.

The content is OK, but should be under "Explanation".

ccremer:
> Why do you remove the navigation structure? I'd like to keep that structure to have it the same on all documentation sites. And it would break the integration into kb.vshn.ch.

I don't think there will be ever any tutorials or how-to pages for an architecture documentation. So everything would be "Explanation" and there we would be limited into which section we can put pages into.

Or did I understand something wrong?

tobru (approved):

PR #2, opened 2021-07-30
Document decision to use Keycloak

ccremer:
My attempt at documenting decisions.
Basically it follows this structure:

* Problem: What are we trying to solve?
* Alternatives: What are our options?
* Decision: What have we decided?
* Rationale: Why did we choose the option?

Additionally, we provide links to:
* Requirements: Relevant requirements documentation.
* Dependent decisions: Some decisions depend on other decisions.

### DIFF
--- a/docs/modules/ROOT/pages/explanation/decisions/keycloak.adoc
+++ b/docs/modules/ROOT/pages/explanation/decisions/keycloak.adoc
@@ -1,0 +1,28 @@
+= Keycloak as IDP
+:apub2: APPUiO Public 2.0
+
+Problem::
+We need to store user (including passwords), team memberships and organizations.
+
+//Relevant requirements::
+//* tbd, links to requirement page(s)
+
+Alternatives::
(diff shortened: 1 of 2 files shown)


ccremer:
(currently this PR contains changes from #1 , should be rebased once #1 is merged)

tobru (changes requested):
Content OK, I'd also fill that under "Explanation".

ccremer:
> I'd also fill that under "Explanation".

I think there will be dozens of documented decisions over the next few iterations. should we all keep them as separate pages? Or make them into one page using sections, or is it possible to create a sub-navigation under "explanation"?

I don't think that "explanation" should contain random bunch of decisions, requirements, diagrams, etc. That would be bad navigation experience since we can clearly group requirements and decisions.

ccremer:
I forgot that indendet navigation is a thing. All good for me.

tobru (approved):

PR #5, opened 2021-07-30
Add system idea

ccremer:
Alternative proposal of #3

### DIFF
--- a/docs/modules/ROOT/pages/explanation/system/idea.adoc
+++ b/docs/modules/ROOT/pages/explanation/system/idea.adoc
@@ -1,0 +1,24 @@
+= System idea
+:apub2: APPUiO Public 2.0
+
+[TIP]
+====
+The system idea displays a high-level architecture of {apub2}.
+====
+
+.{apub2} system idea
+image::system/systemidea.drawio.svg[]
+
+* An instance of Keycloak is responsible for authoritatively store data on:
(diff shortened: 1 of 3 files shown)


ccremer:
Draft: Text doesn't yet explain all components or interactions

docs/modules/ROOT/pages/explanation/system/idea.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+.{apub2} system idea
+image::system/systemidea.drawio.svg[]
+
+* An instance of Keycloak is responsible for authoritatively store data on:
+** Organizations
+** Teams
+** Users
+
+* Data owned by Keycloak will be cached on each Zone.
```
tobru, line 14:
```suggestion
* Data owned by Keycloak will be replicated to each Zone.
```

tobru (approved):
One small rewording, the rest LGTM so far. I'm not sure about everything we write here, but we're in the exploration phase so it's OK to have this currently documented so far. I'm OK going forward with the PR and closing #3.

ccremer:
> I'm not sure about everything we write here, but we're in the exploration phase so it's OK to have this currently documented so far.

Yes, I'm fully aware that certain parts of the diagram are currently "random". It's definitely meant to be further finalized but should be a starting point.

PR #6, opened 2021-07-30
Add decision: Kyverno

ccremer:


docs/modules/ROOT/pages/explanation/decisions/kyverno.adoc
```diff
@@ -0,0 +1,25 @@
...
+  Open Policy Agent is a policy engine written for cloud native environments.
+  Policies are written in Rego, which is declarative query language suited for writing policies.
+
+* https://kyverno.io/[Kyverno].
+  Kyverno is a Kubernetes native policy engine that uses Kyverno-specific CRDs to configure policies.
+
+Decision::
+Kyverno
+
+Rationale::
```
tobru, line 23: Maybe also add a point that Kyverno is able to generate resources which OPA can't.

tobru: Not sure... IMHO we should mention this, even if it solved a different problem and mention it accordingly.

ccremer: It's irrelevant for the policy engine decision. If Kyverno would not be able to generate resources, we'd still choose Kyverno and generate resources with other options.

But I have added the decision for generating resources and linked them together. Is that acceptable?

tobru (approved):

PR #10, opened 2021-08-03
Attempt to document quality requirements

ccremer:
This is my attempt at document quality/non-functional requirements.

The format proposed follows the [SMART](https://en.wikipedia.org/wiki/SMART_criteria) principle:
* S: Specific
* M: Measurable
* A: Attainable
* R: Realizable
* T: Traceable, Time-bound or Testable (literature differs here)

To formulate those, I'm proposing the following structure as scenarios:
* Source: Who  (user or system) triggers an event?
* Stimulus: What happens, what is the event?
* Environment: Where does it happen?
* Artifact: What system component is involved?
* Response: What is the response from the system?
* Response measure: How is the response measured? -> It should be a number, but a yes/no is also acceptable.

For each requirement I have created its own document that can be linked and extended.
An alternative structure would be to document the requirements as a table instead of each document, but linking them would become harder I think.

Each adoc file sits in one of the 8 standardized quality group in [ISO 25010](https://iso25000.com/index.php/en/iso-25000-standards/iso-25010).

### DIFF
--- a/docs/modules/ROOT/pages/references/quality-requirements/performance/resource-quota.adoc
+++ b/docs/modules/ROOT/pages/references/quality-requirements/performance/resource-quota.adoc
@@ -1,0 +1,52 @@
+// According to ISO 25010 (https://iso25000.com/index.php/en/iso-25000-standards/iso-25010)
+// Resource utilization and capacity are in the Performance category.
+= Resource limit per Namespace
+
+Source::
+Developer
+
+Stimulus::
+Deploys new workload in a Namespace
+
+Environment::
+Any APPUiO Zone
+
+Artifact::
+kubectl
+
+Response::
+Workloads are denied if they exceed the resource limit.
+Workloads are the following Kubernetes resources:
+* Pods
+* ConfigMaps
+* Secrets
+* PersistentVolumeClaims
+* Services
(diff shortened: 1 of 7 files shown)


tobru (approved):
Interesting approach, let's see how this progresses and evolves :+1:

docs/modules/ROOT/pages/references/quality-requirements/performance/resource-quota.adoc
```diff
@@ -0,0 +1,52 @@
...
+** Limits: 6 GiB
+* Storage:
+** Request: 50 GiB (total)
+** PersistentVolumeClaims: 10
+* ConfigMaps: 150
+* Secrets: 150
+* Services: 20
+* Images
+** ImageStream Tags: 50 (per ImageStream)
+** ImageStreams: 20
```
ccremer, line 44: These numbers are certainly not set in stone. Like software, they live and breathe in the document.
But to make the requirement measurable, some numbers need to be defined.

corvus-ch (approved):
I kind of like this structure. It certainly is not polished but I guess this is something we get better as we do it. Let's see where this leads us.

Once merged, can you add links to related stories?

PR #12, opened 2021-08-10
Add decisions regarding user names

simu:


docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+* Use the `Email` property of Keycloak user objects.
++
+With this approach, users can't change their E-Mail address after signing up, for the same reasons as they can't change their `Username` property when using that property (see first alternative).
+
+
+Decision::
+* We use the `Username` property of Keycloak user objects as the name for the resulting OpenShift `User` objects
+* We allow users to freely choose their `Username` when signing up
+* We disallow users to change their `Username` after creation of the user in Keycloak
```
simu, line 38: I'm not sure if it makes sense to have all of these in a single document, or if we should split them out

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,57 @@
...
+This is because Openshift's OpenID Connect identity provider uses the OpenID claim `sub` to establish the user's identity, see the https://docs.openshift.com/container-platform/4.8/authentication/identity_providers/configuring-oidc-identity-provider.html[OpenShift documentation].
+The OpenID claim `sub` is mapped to the Keycloak user's `ID` by default.
++
+For all described alternatives, Keycloak can be configured to require that the property must be unique per realm.
+Therefore, we don't risk having OpenShift `User` objects which are associated with multiple Keycloak users with any of the described alternatives.
++
+First, we decide to use the `Username` property as the identifier used for the OpenShift `User` object.
+This choice provides the best trade-off in our eyes, as it allows users to change their E-Mail address at any time and it shows users a human-verifiable output when running `oc whoami`.
++
+Second, we decide to allow users to freely choose their `Username` when signing up.
```
ccremer, line 52: This sentence doesn't provide the reason why we chose it like this.
```suggestion
Second, allowing users to freely choose their `Username` when signing up increases user experience since they can choose a memorable name they like.
```

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,57 @@
...
++
+Third, we decide to disallow users to change their `Username` after creation of the user in Keycloak.
+If a user needs to change their `Username`, we recommend that they register a new user with the desired `Username` and add the new user to the appropriate organizations.
+In the future, we may decide find a solution to allow users to change the `Username` property of their user without having to register a new user.
+One potential approach for such a solution could be to implement extra tooling which fully synchronizes changes to a Keycloak user into the user's associated OpenShift `Identity` and `User` objects.
```
ccremer, line 57: I would not document this here resp. only when we find the need to review our decision. The documentation shouldn't provide insights into the future to avoid making potentially false hopes.

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,57 @@
...
+Therefore, we don't risk having OpenShift `User` objects which are associated with multiple Keycloak users with any of the described alternatives.
++
+First, we decide to use the `Username` property as the identifier used for the OpenShift `User` object.
+This choice provides the best trade-off in our eyes, as it allows users to change their E-Mail address at any time and it shows users a human-verifiable output when running `oc whoami`.
++
+Second, we decide to allow users to freely choose their `Username` when signing up.
++
+Third, we decide to disallow users to change their `Username` after creation of the user in Keycloak.
+If a user needs to change their `Username`, we recommend that they register a new user with the desired `Username` and add the new user to the appropriate organizations.
+In the future, we may decide find a solution to allow users to change the `Username` property of their user without having to register a new user.
```
tobru, line 56:
```suggestion
In the future, we may decide to find a solution to allow users to change the `Username` property of their user without having to register a new user.
```

tobru (approved):

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,57 @@
+= `User` object names in the OpenShift cluster
+
+Problem::
+We need to define an identifier which we use to name users which are synchronized from Keycloak to the OpenShift cluster.
++
+The identifier should have the following properties:
++
+* It should be stable for the lifetime of the user in Keycloak.
+* It must be suitable for use as an OpenShift resource name (namely for the OpenShift `User` resource)
+* It must be used to configure OpenShift `Group` membership when synchronizing groups from Keycloak.
```
ccremer, line 10: I would actually put those requirements into their own documents and refer to them via links.
`It should be stable for the lifetime of the user in Keycloak.` looks to be like requirement independent from any technology, so I guess it goes into the Usability category (this problem is actually independent of Keycloak itself).

The other two are depending/coupled to Keycloak and OpenShift, so I would propose to document those as functional requirements.
Right now, we don't have functional requirements documented yet, but it could be written as a Gherkin scenario that we already use in our Jira Stories. However, for the time being it's fine to leave them as is until we have a good template for functional requirements.

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
ccremer: I kinda agree. The `We use the Username property of Keycloak user objects as the name for the resulting OpenShift User objects` decision makes sense, the other two are separate.

docs/modules/ROOT/partials/nav-explanation.adoc
```diff
@@ -6,3 +6,4 @@
 ** xref:appuio-public:ROOT:explanation/decisions/keycloak.adoc[Keycloak as IDP]
 ** xref:appuio-public:ROOT:explanation/decisions/kyverno-policy.adoc[Kyverno as policy engine]
 ** xref:appuio-public:ROOT:explanation/decisions/kyverno-generator.adoc[Keyverno as resource generator]
+** xref:appuio-public:ROOT:explanation/decisions/usernames.adoc[`User` object names in the OpenShift cluster]
```
ccremer, line 9: I would avoid code markup in navigation links
```suggestion
** xref:appuio-public:ROOT:explanation/decisions/usernames.adoc[User object names in the OpenShift cluster]
```

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
simu: I've added `docs/modules/ROOT/pages/references/quality-requirements/usability/user-arbitrary-name.adoc` (allowing users to choose arbitrary user names is clearly a usability requirement) and `docs/modules/ROOT/pages/references/quality-requirements/usability/user-name-stable.adoc`. However, I'm not sure from what perspective username stability can be considered a usability requirement.

docs/modules/ROOT/pages/references/quality-requirements/usability/user-name-stable.adoc
```diff
@@ -0,0 +1,24 @@
...
+All APPUiO Zones
+
+Artifact::
+IDP (Keycloak)
+
+Response::
+???
+
+Response measure::
+???
```
ccremer, line 19: Something like this?
```suggestion
Developer

Stimulus::
User has desire to change the username.

Environment::
All APPUiO Zones

Artifact::
IDP (Keycloak)

Response::
Username is not changed

Response measure::
* System prevents username change via API with an error message
* UI prevents editing of the username property 
```

docs/modules/ROOT/pages/explanation/decisions/usernames.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+
+Decision::
+We use the `Username` property of Keycloak user objects as the name for the resulting OpenShift `User` objects
+
+Rationale::
+
+Regardless of the choice of identifier to determine how the OpenShift `User` object is named, the user's identity in the OpenShift cluster is always tied to their Keycloak user's `ID` property.
+This is because Openshift's OpenID Connect identity provider uses the OpenID claim `sub` to establish the user's identity, see the https://docs.openshift.com/container-platform/4.8/authentication/identity_providers/configuring-oidc-identity-provider.html[OpenShift documentation].
+The OpenID claim `sub` is mapped to the Keycloak user's `ID` by default.
```
ccremer, line 40: I would move this to the supporting information in the Problem section, as it doesn't directly influence the decision.

ccremer (approved):

corvus-ch (approved):

PR #14, opened 2021-08-11
Add architecture reference docs for authentication

simu:


simu:
I'm not sure if the structure, location and naming makes sense. Naming is currently probably still a mix between "APPUiO Public 2.0 zone" and "OpenShift cluster".

docs/modules/ROOT/pages/references/architecture/openshift-idp-data-model.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+{apub} users are represented as a pair of `User` and `Identity` objects in each {zone}.
+These objects are created when a user first logs in on a {zone}.
+The user's Keycloak property `Username` is used to name the `User` object in each {zone}.
+
+{apub} organizations are represented as `Group` objects in each {zone}.
+The `Groups` are synchronized from Keycloak to all {zone}s in regular intervals.
+The group synchronization is implemented with the RedHat Communities of Practice https://github.com/redhat-cop/group-sync-operator[`group-sync-operator`].
+
+Organization memberships are represented in each {zone} by adding the user's `User` object name to the organization's `Group` object.
```
ccremer, line 14: I think it makes sense to create a diagram for this data model. Should be easily doable with Drawio or so.

docs/modules/ROOT/pages/references/architecture/keycloak-data-model.adoc
```diff
@@ -0,0 +1,15 @@
+= Keycloak data model
+
+:apub: APPUiO Public 2.0
```
tobru, line 3: What's the reason for that as it is duplicated in each page anyways? Wouldn't it make more sense to have this configure globally for all pages? Afaik OpenShift does it that way for the OpenShift documentation, but I don't know how.

tobru (approved):
LGTM, only a non-blocking comment and the missing diagram.

docs/modules/ROOT/pages/references/architecture/keycloak-data-model.adoc
simu: This attribute is now defined globally in `docs/antora.yml`.

corvus-ch (approved):

tobru (approved):

ccremer:
I have pushed commits that adds the "global" data model, whereas the "identity data model" shows how just User and Organization entities are implemented.

PR #16, opened 2021-08-17
Rename into APPUiO Cloud

tobru:
We decided to rename "APPUiO Public 2.0" into APPUiO Cloud.


simu (approved):
LGTM, I didn't check if you missed an "APPUiO Public 2.0" somewhere.

corvus-ch (approved):

PR #18, opened 2021-08-25
Define naming scheme for domains and zones

corvus-ch:
## Checklist


megian:
I think we require some variables in the name:
* Geographic Location (Must) -> Rationale: Customer's often have legal constraints where data is at rest or processed 
* Availability Zone (Must) -> Rationale: It's important for DR/BCM concepts to understand the geographic location.
* Additional identifier (Must) -> if there are multiple independent clusters
* Cloud provider (Optional)

I would prefer for the zone to take something people already aware of. The concept of <region> and <availability zone> is very common in the hyper-scaler world.
Probably a combination like https://service.unece.org/trade/locode/ch.htm or https://rook.io/docs/rook/v1.7/ceph-cluster-crd.html#osd-topology could be reasonable. I would prevent taking the zone naming from the specific cloud provider. This causes a very inconstant naming, not predictable in my head.

Additional identifier is required, because there might be additional requirements for a cluster at the same location. In example there might be clusters with a higher availability, higher security constraints like SwissOps only or just a newer instance of the same like OpenShift 5.x.

corvus-ch:
> I think we require some variables in the name:
> 
> * Geographic Location (Must) -> Rationale: Customer's often have legal constraints where data is at rest or processed
> * Availability Zone (Must) -> Rationale: It's important for DR/BCM concepts to understand the geographic location.
> * Additional identifier (Must) -> if there are multiple independent clusters
> * Cloud provider (Optional)
> 
> […]

Do we realy need that information to be encoded in to pieces of a domain? I would argue, that this is better covered by documentation.  The name is then just a human memorable identifier.

akosma (approved):

corvus-ch:
I wait with the rationale of the zone nameing until we have that settled. I sense heavy discussion ahead.

docs/modules/ROOT/pages/explanation/decisions/dns-nameing.adoc
```diff
@@ -0,0 +1,39 @@
+= DNS nameing scheme
```
TheBigLee, line 1: Typo: Should be `naming`and not `nameing`

docs/modules/ROOT/pages/explanation/decisions/zone-nameing.adoc
```diff
@@ -0,0 +1,30 @@
...
+* Using a randomly generated identifier like `rapid-sound-0933` or `rapid-sound` (`rapid-sound.appuio.cloud`).
+  We have tried that already with https://syn.tools[Project Syn].
+  This is way better than for example a random string of characters or a UUID.
+  From the perspective of daily operation, we have learned that this was still too much mental load even for a small number of items.
+* Use top level domain to encode country locality (`<zone>.appuio.ch`, `<zone>.appuio.de`).
+  This does not work as not every country has a country specific top level domain (for example the United States of America).
+* Use the cloud provider followed by a numerical or alphabetical index (`cloudscale-a.appuio.cloud` or `exoscale-1.appuio.cloud`).
+* Use the country code followed by a numerical or alphabetical index (`ch-1.appuio.cloud` or `de-a.appuio.cloud`).
+
+Decisison::
```
TheBigLee, line 17: Typo: Should be `Decision`

docs/modules/ROOT/pages/explanation/decisions/dns-nameing.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+Problem::
+{product} is a web based service offering.
+As such it is accessible through several domains.
+We need a common understanding how we name things and what domains to use for making it accessible.
+
+Alternatives::
+* In the past, we used different domains for things managed by us and things deployed by users. This due to a concern regarding cross domain cookie access.
```
TheBigLee, line 9: The last sentence sounds strange I would rewrite it like this:
`... things deployed by users, due to a concern ...

docs/modules/ROOT/pages/explanation/decisions/dns-nameing.adoc
```diff
@@ -0,0 +1,39 @@
...
+Problem::
+{product} is a web based service offering.
+As such it is accessible through several domains.
+We need a common understanding how we name things and what domains to use for making it accessible.
+
+Alternatives::
+* In the past, we used different domains for things managed by us and things deployed by users. This due to a concern regarding cross domain cookie access. 
+
+
+Decisison::
```
TheBigLee, line 12: Typo: Should be `Decisions`

docs/modules/ROOT/pages/explanation/decisions/dns-nameing.adoc
```diff
@@ -0,0 +1,39 @@
...
+The base domain will be `appuio.cloud`.
+On the third level, we will serve services that are zone independent.
+Services bound to specific zone, will be served through the fourth level where the third level will be the zone's identifier.
```
TheBigLee, line 15: either:
*Services bound to **a** specific zone
*Services bound to specific zone**s**

docs/modules/ROOT/pages/explanation/decisions/zone-nameing.adoc
```diff
@@ -0,0 +1,30 @@
...
++
+The zone identifier will build with the following the patter: `<provider>-<region>-<index>`.
```
TheBigLee, line 19: should be: will **be** build

megian:
> Do we realy need that information to be encoded in to pieces of a domain? I would argue, that this is better covered by documentation. The name is then just a human memorable identifier.

I've not a strong opinion about that. Choosing an alias identifier and do a mapping in the documentation is also a valuable option to me. This could be either an random name (small-wood-4148 or cluster-z5d7e). However more people voted for having the attributes explicit in the name. From the operation work with such a naming, I can just tell that using random name is against "They can also be typed manually without being a big hurdle." and "As they are following a simple pattern, a versed person should be able to guess them.". There is always a virtual map in your head and this causes overhead and increases the potential something is applied on the wrong cluster because of a wrong human mapping.

Concatenate known patterns to a name is most probably easier for our human brain. The less the better.

Finally: At least there must be an unique identifier and there might be multiple clusters with the same parameters.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
```diff
@@ -0,0 +1,34 @@
...
+`<provider>-<region>-<index>`
+****
++
+`<provider>`::: Name of the cloud provider the cluster is hosted on (`cloudscale`, `exoscale`, `aws`, `azure`, `gcp`, etc.).
+`<region>`::: Region name, taken verbatim from the cloud providers own naming scheme (`rma`, `ch-dk-2`, `eu-central-1`, `switzerlandnorth`, `europe-west6`).
+`<index>`::: Zero indexed numerical identifier.
+
+Examples:
+
+* `cloudscale-rma-0.appuio.cloud` first cluster hosted on Cloudscale and placed in the region located in Rümlang.
```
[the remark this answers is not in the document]
corvus-ch: By my understanding, regions are geographical locations and zones are for separation of failure. Given a provider does provide multiple zones, spreading the cluster is an option and persanlly, I would prefer to do so.

I will make a note to clarify on the fuzziness of Cloudscale as not everybody is aware that they implicilty have regions and zones.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
```diff
@@ -0,0 +1,34 @@
...
+* `exoscale-ch-dk-2-a.appuio.cloud` third cluster hosted on Exoscale and placed in the region located in Zürich.
```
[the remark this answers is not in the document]
corvus-ch: I was playing around with both. I am still not undecided which one is better.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
```diff
@@ -0,0 +1,34 @@
...
+Decision::
+The zone identifier will be built following the pattern:
++
+****
+`<provider>-<region>-<index>`
+****
++
+`<provider>`::: Name of the cloud provider the cluster is hosted on (`cloudscale`, `exoscale`, `aws`, `azure`, `gcp`, etc.).
+`<region>`::: Region name, taken verbatim from the cloud providers own naming scheme (`rma`, `ch-dk-2`, `eu-central-1`, `switzerlandnorth`, `europe-west6`).
+`<index>`::: Zero indexed numerical identifier.
```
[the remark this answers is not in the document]
corvus-ch: That is way I brought up in the examples. Reading it from you, made me tend to the alphabetical index.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
```diff
@@ -0,0 +1,46 @@
...
+====
+
+[NOTE]
+====
+On Cloudscale, you will most likely only see `rma1` and `lpg1`.
+One is inclined to assume those are the regions, but actually, those are the zones.
+As Cloudscale only has one zone per region, it is easy to confuse one for the other.
+====
+
+== Rationale
```
ccremer, line 46: I'm a bit unhappy with the formulation/structure of this decision.

The way Alternatives is initially meant to be structured is that they don't contain random partial ideas, but actual, complete proposals where one could choose one of. The naming scheme described currently in the Decision section isn't listed at all in the Alternatives section but instead combines several ideas from the Alternatives into one. The Decision section could just contain one line: Which alternative was choosen. Probably makes sense to label the Alternatives or at least number them, so we can say "We've choosen Option 2"

The WHY belongs to Rationale which is currently completely empty. 

Examples of how Cloudscale naming schemes could look like belong to the Alternatives IMO.

For example, listing `Use top level domain to encode country locality (<zone>.appuio.ch, <zone>.appuio.de)` isn't a complete Proposal as the values of `<zone>` is undefined.

Options that were never really on the table should be removed completely. I'd rather have 2 complete Alternatives than 10 ideas.

docs/modules/ROOT/pages/explanation/decisions/dns-naming.adoc
```diff
@@ -0,0 +1,42 @@
...
+====
+* `appuio.cloud` and `www.appuio.cloud` are an exception to the above rule and will be used to host the marketing homepage for {product}.
+* `id.appuio.cloud` The identity provider for {product}.
+* `console.appuio.cloud` A web application to manage cross zone concerns.
+* `console.<zone>.appuio.cloud` The OpenShift console of a zone.
+* `<route>.apps.<zone>.appuio.cloud` A route hosted on a zone relying on the zones default app domain.
+====
+
+The content of `<route>` is covered by an OpenShift core feature and will consist of the name of the route object and the namespace it is located in.
+What goes into `<zone>` is covered by xref:appuio-cloud:ROOT:explanation/decisions/zone-naming.adoc[Zone naming scheme].
```
ccremer, line 30: Similar to my other comment, the Alternatives section should contain complete proposals, whereas the Decision section only referring to which Alternative choosen.

So I'd move the content of the Decision section to an Alternative. And then add an example for a sentence of "In the past, we used different domains for applications managed by us and applications deployed by users, due to a concern regarding cross-domain cookie access." -- I don't quite understand this. How did that look like?

ccremer (changes requested):

docs/modules/ROOT/pages/explanation/decisions/dns-naming.adoc
```diff
@@ -0,0 +1,42 @@
...
+Services bound to a specific zone, will be served through the fourth level where the third level will be the zone's identifier.
+Apps served through a zone's default domain use the subdomain `apps` of a zone's base domain.
+
+.List of examples
+====
+* `appuio.cloud` and `www.appuio.cloud` are an exception to the above rule and will be used to host the marketing homepage for {product}.
+* `id.appuio.cloud` The identity provider for {product}.
+* `console.appuio.cloud` A web application to manage cross zone concerns.
+* `console.<zone>.appuio.cloud` The OpenShift console of a zone.
+* `<route>.apps.<zone>.appuio.cloud` A route hosted on a zone relying on the zones default app domain.
```
corvus-ch, line 26: ingress vs route

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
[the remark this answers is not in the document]
megian: What speaks pro alphabetical is that there are 26 characters instead of 10 for one single char. Also the sort algorithm is less issues with. But since we have this random in the cloud provider naming as well, I think at the end it doesn't matter so much. Open to both.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
```diff
@@ -0,0 +1,34 @@
...
+{product} will consist of several clusters we refer to as zones.
+We need a common understanding how we name them.
+
+Alternatives::
+* Using a randomly generated identifier like `rapid-sound-0933` or `rapid-sound` (`rapid-sound.appuio.cloud`).
+  We have tried that already with https://syn.tools[Project Syn].
+  This is way better than for example a random string of characters or a UUID.
+  From the perspective of daily operation, we have learned that this was still too much mental load even for a small number of items.
+* Use top level domain to encode country locality (`<zone>.appuio.ch`, `<zone>.appuio.de`).
+  This does not work as not every country has a country specific top level domain (for example the United States of America).
```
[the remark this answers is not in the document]
megian: The only downside of having everything on one zone is, that if we break DNS somehow, this can have an effect to all our appuio clusters. Would not change it, but might be we should document this.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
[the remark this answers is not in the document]
ccremer: The sorting really depends on the algorithm. Even with alphabetical characters, system sort "unexpectedly":
```
a
aa
aab
b
bb
```
So, in the unlikely event we have more than 26 clusters in a region, "aa" would still be listed second, even if it's actually the 27th cluster. Like sorting of numbers:
```
1
11
2
```
So, without pre-zeroes like `01, 02`, I would just ignore the sorting of the indexes...

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
ccremer: As discussed in-person, we will not document the naming scheme as an architecture decision since the naming scheme has no relevant impact on the architecture.
Instead we simply document it as a definition in the reference.

docs/modules/ROOT/pages/explanation/decisions/dns-naming.adoc
ccremer: See other comment

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
```diff
@@ -0,0 +1,99 @@
...
+|===
+| Hostname | Description
+
+|`appuio.cloud`, `www.appuio.cloud`
+| Hosting the marketing homepage for {product}
+
+|`id.appuio.cloud`
+|identity provider for {product}
+
+|`console.appuio.cloud`
```
megian, line 23: " console" might be confusing because it interferes with OpenShift console. I would prefer something a different name to make it clear, this are different "consoles". Like control, portal, my.

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
```diff
@@ -0,0 +1,99 @@
...
+Apps served through am {zone}'s default domain use the subdomain `apps` of a zone's base domain.
+
+.Known hosts and schemes of 'appuio.cloud'
+|===
+| Hostname | Description
+
+|`appuio.cloud`, `www.appuio.cloud`
+| Hosting the marketing homepage for {product}
+
+|`id.appuio.cloud`
```
megian, line 20: Probably it would make sense to be more specific here. `keycloak.appuio.cloud` and a CNAME for specific protocols like OCID, SAML. This would it make more easy, if we might switch to a different solution than Keycloak in the future.

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
```diff
@@ -0,0 +1,99 @@
...
+| `console.<zone>.appuio.cloud`
+| OpenShift console frontend of an {zone}
+
+| `api.<zone>.appuio.cloud`
+| Kubernetes API endpoint (e.g. for KubeConfig)
+
+| `cname.<zone>.appuio.cloud`
+| The DNS entry where users can point CNAME entries to
+
+| `<hostname>.apps.<zone>.appuio.cloud`
```
megian, line 35: `<app>.apps.<zone>.appuio.cloud`? because it's actually no host in a classical sense.

docs/modules/ROOT/pages/explanation/decisions/zone-naming.adoc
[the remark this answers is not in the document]
megian: As I said. I do not have a strong opinion here. Fair enough just to keep it as is.

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
ccremer: Switching to another technology than Keycloak is a great architectural change. I don't think we should already prepare for that.
Adding CNAMEs like OCID and SAML would probably be purely technical implementation detail, not sure about the relevance in the architecture documentation.

megian (approved):

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
tobru: I'd rather not add such a complexity for a migh-be-probably-maybe feature.

docs/modules/ROOT/pages/references/dns-naming-scheme.adoc
tobru:
```suggestion
|`my.appuio.cloud`
```

I agree with the reasoning. Let's use this suggestion for now. As we don't know yet the details about this application we can later change the URL again when we know more... We're still weeks / months away from having something to deploy anyways.

tobru (approved):

tobru:
IMHO it's good enough for now and can be merged as it is now. Thanks for all the good discussion and input.

PR #23, opened 2021-09-14
Decision about source of Users

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanation/decisions/source-of-users.adoc
+++ b/docs/modules/ROOT/pages/explanation/decisions/source-of-users.adoc
@@ -14,6 +14,7 @@ Alternatives::
 
 Decision::
 The User base won't be in {idp}, it has to be in the resellers IdP, e.g. in VSHNs IdP.
+It has to be made available to {idp} using xref:how-to/keycloak-brokering.adoc[identity brokering].
 
 Rationale::
 The User base belongs to the reseller and not to APPUiO Cloud.
--- a/docs/modules/ROOT/pages/explanation/decisions/source-of-users.adoc
+++ b/docs/modules/ROOT/pages/explanation/decisions/source-of-users.adoc
@@ -1,0 +1,20 @@
+= Source of Users
+
+Problem::
+The source of the Users (User base) has to be somewhere well-defined.
+
+Relevant decisions::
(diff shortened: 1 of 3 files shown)


tobru:
Before merging this we might have to revisit some linked relevant decisions as they are sometimes conflicting.

docs/modules/ROOT/pages/explanation/decisions/source-of-users.adoc
```diff
@@ -0,0 +1,21 @@
...
+Alternatives::
+* Store user base in xref:appuio-cloud:ROOT:explanation/decisions/keycloak.adoc[{idp}].
+
+Decision::
+The User base won't be in {idp}, it has to be in the resellers IdP, e.g. in VSHNs IdP.
+It has to be made available to {idp} using xref:how-to/keycloak-brokering.adoc[identity brokering].
+
+Rationale::
+The User base belongs to the reseller and not to APPUiO Cloud.
+The {idp} only stores a copy of the User and enhances it with APPUiO Cloud specific annotations.
```
ccremer, line 21: * I would avoid any mentions to VSHN at all
* If there is only 1 proposal, what is there to decide about? -> The decision should only name which alternative was chosen, not describe the proposal.
```suggestion
== Alternatives

[qanda]
Store user base in xref:appuio-cloud:ROOT:explanation/decisions/keycloak.adoc[{idp}]::
--> more details maybe?

The User base won't be in {idp}::
It will be in the reseller's resp. cluster-owner's IdP, made available to {idp} using xref:how-to/keycloak-brokering.adoc[identity brokering].

Decision::
The User base won't be in {idp}

Rationale::
The User base is provided by the reseller and not by {product}.
The {idp} only stores a copy of the User and adds {product} specific annotations.
```

ccremer (changes requested):
tbh, I'm not sure this is a decision we need to document. If I remember correctly, we discussed that the _intention_ or _most common use case_ is identity brokering because we as VSHN and other current customers would use it that way. 
But this doesn't exclude the fact that the APPUiO Keycloak _can_ be used as the primary IdP if some reseller doesn't have its own IdP. Or at least I don't see a reason why there has to be _always_ ID brokering in place.

This decision doesn't have an impact on the architecture (I don't see one). So maybe we can just somewhere better describe the intended (but not strictly required) usage?

tobru:
After some thoughts and together with #24 I decide to close this PR again, it's not really needed.

PR #35, opened 2021-09-27
Framework pattern for Control API

tobru:
* The Control API is the user-facing API that enables users to orchestrate APPUiO Cloud, but also query for entities from 3rd party providers (e.g. ERP system)
* This decision lays the foundation how this API is going to look like.
* TLDR: Kubernetes API & Controllers

## Checklist


docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+It must be possible to connect various different backend systems which implement the same functionality so that it is a pluggable system and can be re-used by other parties, which might have different backends for the same task.
+{product} API only specifies the interface, the implementation is then user specific.
+
+== Relevant requirements
+
+* xref:references/quality-requirements/maintainability/portal-api-modularity.adoc[Portal API Modularity]
+* xref:references/quality-requirements/reliability/portal-api-data-recoverability.adoc[Portal API Data Recoverability]
+* xref:references/quality-requirements/performance/portal-api-data-freshness.adoc[Portal API Data Freshness]
+
+== Proposals
```
ccremer, line 19: The proposals are lacking details for the unknowing user. I'd say we support the proposals with the diagrams (e.g. the one that is currently in External-Services page) and a new one that better displays the interaction between Kubernetes, the user and the 3rd party system.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+Persistent Storage::
+Storing data is built-in directly in the core.
+We can store temporary data without having to implement a storage backend.
+
+API Extensibility::
+The API can easily be extended by leveraging `CustomResourceDefinition` (CRDs).
+
+Scalability and Availability::
+It's a core feature of the Kubernetes API server to be highly scalable.
+High-availability possibilities are directly built-in.
```
ccremer, line 58: This is true for both proposals, since with Ingress we can also make REST API HA. So I would remove it since it has no impact. Unless there's an important aspect in how this HA is achieved, but I don't see one in either case.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+API definition vs. Implementation::
+By using `CustomResourceDefinition` (CRDs) to define the API structure, we can easily decouple the interface and the implementation.
+The Portal API client uses the standardized Portal API definition, whereas the implementation in the background communicates with the specific implementation of the backend system.
```
ccremer, line 62: The decoupling of interface from implementation is also true for both proposals, the rationale doesn't favor one over the other.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+CLI client::
+We get a CLI client "for free".
+The Kubernetes CLI client `kubectl` can display and work with the custom API endpoints of the Portal API out-of-the-box.
+Think of `kubectl get invoices`.
+
+Implementation boilerplate with kubebuilder::
+For the implementation of the CRD-defined interface, there already exists a good ecosystem, like kubebuilder, Operator SDK or crossplane-runtime.
```
ccremer, line 70: If the other proposal were REST-like (which it is in my mind, just not written down anywhere), then there is also a great ecosystem around web frameworks, authentication, client-generators etc. So that would be true for both proposals thus irrelevant.
However, I would argue that the company behind APPUiO Cloud (VSHN obvi) has more experience with Kubebuilder and operators than REST applications, so that would be a better reason IMO.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Frontend Development and Testing::
+For the implementation and testing of the Portal API frontend, the backend implementation doesn't actually need to exist.
+Data can be faked by just creating test-objects, defined by the custom CRDs, until the actual implementation of the controller is done ("Fake it 'till you make it").
+It also allows for having a simple test environment.
```
ccremer, line 75: Faking is also easily possible with a dummy REST backend. Granted, the effort is more, but "Data can be faked by just creating test-objects, ..., until the actual implementation ... is done" alone holds true for both, so that reason doesn't really favor one over the other.
I would only mention that the effort for a test environment is easier with Kubernetes/CRD than a dummy REST backend.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Experience from Project Syn::
+We already got experience implementing such an architecture from Project Syn.
```
ccremer, line 78: I would avoid "We" whenever possible
```suggestion
Experience from Project Syn::
Project Syn features a similar architecture which serves as inspiration.
```

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+High-availability possibilities are directly built-in.
+
+API definition vs. Implementation::
+By using `CustomResourceDefinition` (CRDs) to define the API structure, we can easily decouple the interface and the implementation.
+The Portal API client uses the standardized Portal API definition, whereas the implementation in the background communicates with the specific implementation of the backend system.
+
+CLI client::
+We get a CLI client "for free".
+The Kubernetes CLI client `kubectl` can display and work with the custom API endpoints of the Portal API out-of-the-box.
+Think of `kubectl get invoices`.
```
ccremer, line 67: These sentences as they are now currently only describe a feature of the proposal, so that should belong to the proposal itself.
Also, there's currently no requirement written down (or mentioned in the Problem section) that would require a CLI in the first place.

Either move these sentences to the proposal, or highlight the advantage that there's no additional effort needed to implement a CLI compared to a REST API (even though a CLI is currently only a future-music feature).

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Asynchronous::
+It allows us to build upon an asynchronous architecture which when working with multiple backend systems will prove to be more stable and easier to implement.
+
+Persistent Storage::
+Storing data is built-in directly in the core.
+We can store temporary data without having to implement a storage backend.
+
+API Extensibility::
+The API can easily be extended by leveraging `CustomResourceDefinition` (CRDs).
```
ccremer, line 54: This is not really a reason that favors one over the other. I mean, REST APIs can also be extended. I would either remove it or move to the proposal

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Authorization::
+We can make use of standard Kubernetes RBAC access control and could even enhance it with a policy engine like Kyverno.
+
+Asynchronous::
+It allows us to build upon an asynchronous architecture which when working with multiple backend systems will prove to be more stable and easier to implement.
+
+Persistent Storage::
+Storing data is built-in directly in the core.
+We can store temporary data without having to implement a storage backend.
```
ccremer, line 51: This is actually more of a disadvantage of the Kubernetes API approach. With the REST approach, there's simply no need for persistent storage, while with Kubernetes there's the need for a HA and backed up etcd (and maybe PVs).
(an implementation may need persistent storage, but that's not a problem of APPUiO Cloud)

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Authentication::
+We can leverage what we already use for the {zone}s.
+Authentication is built-in to the Kubernetes API server and it's even pluggable.
+
+Authorization::
+We can make use of standard Kubernetes RBAC access control and could even enhance it with a policy engine like Kyverno.
+
+Asynchronous::
+It allows us to build upon an asynchronous architecture which when working with multiple backend systems will prove to be more stable and easier to implement.
```
ccremer, line 47: The argument of "easier to implement" is probably debatable. It's often enough the case that Asynchronous applications have to solve an entire different set of problems than synchronous operations (strong vs. eventual consistency), yet those problems aren't necessarily "easier".
I would rephrase it so that we highlight the fact that we have more experience with eventual-consistency-like architecture by working everyday with Kubernetes.

ccremer (commented):
The rationale reads like a marketization/promotion for Kubernetes, but we don't need to do that. The rationale is intended to highlight why the decision more appropriate than all other proposals.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: Does it make sense to differentiate the Kubernetes proposal between plain Operator/CRD and one that uses Crossplane under the hood?
Because implementing the proposal with Crossplane adds a technology dependency to Crossplane which is important to mention IMO, worthy of its own decision if we don't mention it as a separate proposal.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Removed

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: What should I do with this comment? How is that one linked to the linked sentence?

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Amended

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Amended

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Removed

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Removed

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Amended

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Moved to proposal

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: I'll work out some diagrams later :+1: 

I wouldn't differentiate which framework is used to implement the controller, as it basically all boils down to CRDs, even with Crossplane. The main point of this decision is the use of the Kubernetes API as the contract, the exact implementation of the controller is left to the user of it.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: While it's true that REST APIs can also be extended, it's so much simpler to do with the Kubernetes API that I'd like to keep it here to emphasize this. Extensibility is a core feature of the Kubernetes API.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: For me that's a clear advantage of the Kubernetes API, it's there "for free", no need to use other third party systems when needed (e.g. Redis). Also, with the requirement in #37, there is nothing stored which needs a backup.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Good input, amended.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
corvus-ch: > The main point of this decision is the use of the Kubernetes API as the contract, the exact implementation of the controller is left to the user of it.

Agreed, however, it might be worth to list those potential ways of implementation. Mentioning them, would make this less abstract to read.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,73 @@
...
+Reuse the Kubernetes API to leverage all the core-plumbing which already exists: Authentication, RBAC, scalability, well-defined resource-model, extensibility with `CRDs`, aggregated API server and much more.
+This seems to be a common topic in the Cloud Native world, see for example the experiment called https://github.com/kcp-dev/kcp[kcp - a minimal Kubernetes API server]. We could get a CLI client "for free" if we wanted. The Kubernetes CLI client `kubectl` can display and work with the custom API endpoints of the Portal API out-of-the-box.
+
+== Decision
+
+Kubernetes API
+
+== Rationale
+
+The Kubernetes API server is a perfect match for this use-case. With it, we can many core API plumbing features "for free":
```
corvus-ch, line 37:
```suggestion
The Kubernetes API server is a perfect match for this use-case.
With it, we get many core API plumbing features "for free":
```

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,73 @@
...
+VSHN has already a lot of experience writing Kubernetes-based controllers and knows how to work with the Kubernetes API.
+
+Frontend Development and Testing::
+For the implementation and testing of the Portal API frontend, the backend implementation doesn't actually need to exist.
+Data can be faked by just creating test-objects, defined by the custom CRDs, until the actual implementation of the controller is done ("Fake it 'till you make it").
+It also allows for having a simple test environment.
+This would also hold true for the custom API implementation, but would be much more engineering effort.
+
+Experience from Project Syn::
+Project Syn features a similar architecture which serves as inspiration.
```
corvus-ch, line 73: I am missing a few things.

* Client for free (yes it is mentioned above but I would again like to see on this description list)
* Fake it until you make it: the ability to work with the API from the client perspective withough having to write custom code (not counting CRDs here).

corvus-ch (changes requested):

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: Learning from the K8up v2 case (with renamings, adding and removing fields etc.) we both know that this extensibility is not so easy in practice because CRDs are both API specs AND data/document record at the same time.
With that argument, I'd say that a REST API approach has the advantage that it doesn't need to deal with persistence since it's abstracted away. The extensibility from Kubernetes is more limited because of the data records being directly attached to spec.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: Hm, right, with #37 no persistent storage is required for either proposal. So maybe completely remove it?

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: That's not really true, the Kubernetes API would offer a lot of tooling to deal with changing API specs and converting them between API versions. The API versioning plumbing the Kubernetes API brings is very well-thought-out. What we did in K8up v2 is to actually rename the whole API, not just changing spec or version.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Client for free was on this list, but @ccremer mentioned it to be moved away because it isn't listed as requirement. What should I do?
"Fake it until you make it" is already there under "Frontend Development and Testing".

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: Well, still. In case of data scheme changes (the spec in CRD) that change needs to be populated into the clients as well (e.g. frontend, CI/CD systems etc.). With both approaches, only non-breaking extensions can be transparently added.
So, considering that in both approaches breaking changes need changes in all clients, I think this argument is basically redundant and I would simply remove it :shrug: 

What is the exact reason why you think extensions in Kubernetes is "simpler" to do than with REST?

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Updated the proposal accordingly.

tobru:
Rephrased the "Rationale" part, moved the most bunch to "Additional information" and reworded it to not use "we".

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Because the Kubernetes API is already prepared and _a lot of thought_ has been put into all the concepts of extensibility. As well there is a lot of plumbing around to assist with changing API specs and gradually moving to breaking changes of the API.

This could all be done with a custom API as well, but why should we want to re-invent the wheel of this "boring" stuff again when others already solved it for us so that we can focus on the actual case?

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Habemus diagram now :+1:

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,103 @@
...
+Custom API implementation::
+Implementation of a custom (REST or GraphQL) API which connects to all the backend system and provides the abstraction to them.
+This can be a synchronous or asynchronous connection, depending on the implementation.
+Authentication, authorization and other core API features will have to be implemented specific to this API.
+
+Kubernetes API::
+Use the Kubernetes API server to leverage all the core-plumbing which already exists there: Authentication, RBAC, scalability, well-defined resource-model, extensibility with `CRDs`, aggregated API server and much more.
+This seems to be a common topic in the Cloud Native world, see for example the experiment called https://github.com/kcp-dev/kcp[kcp - a minimal Kubernetes API server].
++
+A CLI client could be there "for free" if wanted. The Kubernetes CLI client `kubectl` can display and work with the custom API endpoints of the Portal API out-of-the-box.
```
corvus-ch, line 33:
```suggestion
A CLI client could be there "for free" if wanted.
The Kubernetes CLI client `kubectl` can display and work with the custom API endpoints of the Portal API out-of-the-box.
```

corvus-ch (approved):

ccremer:
I've allowed myself to rebase to latest master branch and rephrasing the proposal.
Basically I've extended the idea for both sides, but removed marketing and judgemental language

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,103 @@
...
+====
+
+== Decision
+
+Kubernetes API
+
+== Rationale
+
+The Kubernetes API approach has been chosen because the engineers at VSHN have a lot of know-how working with it.
+It allows to start fast as the important API plumbing is provided right from the beginning from the Kubernetes API server.
```
tobru, line 80:
```suggestion
It allows to start fast as the important API plumbing is provided right from the beginning from the Kubernetes API server.
And it adheres to the https://handbook.vshn.ch/hb/beliefs.html#_beliefs[VSHN Company Beliefs], mainly the part "For that we use".
```

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
```diff
@@ -0,0 +1,103 @@
...
+Operations can be a synchronous or asynchronous.
+
+With a well-defined interface, clients can be generated for all major programming languages, for example for frontends or CLI apps.
+
+image::decision/rest-api-proposal.drawio.svg[]
+
+.Invoicing example
+[example]
+====
+. At the end of a month, the Portal API collects metrics to prepare a new invoice.
```
tobru, line 37: That's not what the API is used for. This task is part of the to-be-defined metering/billing system.
Can we find a better example?

ccremer: I understood the `api.appuio.cloud` to be the business logic or orchestration layer for all things, not just the frontend...

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: I removed this sentence because it's true in both cases. The REST API would also be running on Kubernetes...

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: Nope, that's not the idea of the API. The API is the interconnecting point between external systems. The whole architecture is meant to be implemented in microservices.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
tobru: I think it's important to link to the Company Beliefs as this is an important reason for the decision.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: I'm sorry, here we have entirely different understandings of "API". I don't see any requirement saying or justifying "it has to be microservices". I don't see a need to "enforce" the complexity of microservices architecture unless there are good reasons for it.
I think we need to define new terms (not using "API" please) in the glossary and agree on the same language first before discussing this further.

docs/modules/ROOT/pages/explanation/decisions/portal-api.adoc
ccremer: I've integrated this into the problem now, making it a precondition for any proposal.

ccremer:
Hopefully last round :)
I've rebased onto #39 , please don't get confused by the diff ;)

docs/modules/ROOT/pages/explanation/decisions/control-api.adoc
```diff
@@ -0,0 +1,130 @@
...
+. A foreign ERP plugin queries the metrics database to generate new invoices.
+. Eventually, the foreign ERP system has generated the invoice along with a PDF version.
+. Clients requesting a list of invoices will transparently query the ERP plugin which will then return a set of relevant invoices.
+====
+
+=== Kubernetes API
+
+The Kubernetes API server provides a standardized API pattern that comes with authentication, RBAC, well-defined resource-model or extensibility with `CRDs`.
+By design, Kubernetes is a API framework that works with reconciling a desired state.
+Business logic or orchestration is asynchronous by nature and this leads to an evetual-consistent behavior of the platform.
```
tobru, line 69:
```suggestion
Business logic or orchestration is asynchronous by nature and this leads to an eventual-consistent behavior of the platform.
```

tobru:
Now it's straight to the point! Thank you very much @ccremer for the effort here, for me it's now a very clear: LGTM :rocket:

corvus-ch (approved):
Please remember to first merge #39 and then rebase this one before merge.

ccremer (approved):

PR #36, opened 2021-09-27
Portal API data freshness requirement

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/quality-requirements/performance/portal-api-data-freshness.adoc
+++ b/docs/modules/ROOT/pages/references/quality-requirements/performance/portal-api-data-freshness.adoc
@@ -1,0 +1,27 @@
+= Portal API Data Freshness
+
+Source::
+Portal API client
+
+Stimulus::
+Portal API client reads data from Portal API
+
+Environment::
+{product} Portal API client
+
+Artifact::
+Portal API client and Portal API
+
+Response::
+Portal API returns the requested data
+
+Response measure::
+The read data is at most 10 minutes older than the original data in the backend system (source-of-truth)
+
+Rationale::
+The data in the API is not guaranteed to be up-to-date to the backend system because of its asynchronous nature.
+For the data which is presented by the API it doesn't matter when the data lacks a bit behind.
+This circumstance has to be communicated to the user of the API by documentation and a data freshness indicator.
+
+See also::
+* xref:references/quality-requirements/reliability/portal-api-data-recoverability.adoc[Portal API Data Recoverability]
--- a/docs/modules/ROOT/partials/nav-reference.adoc
+++ b/docs/modules/ROOT/partials/nav-reference.adoc
@@ -16,6 +16,7 @@
 *** xref:appuio-cloud:ROOT:references/quality-requirements/performance/ns-create-time.adoc[Time to create Namespace]
 *** xref:appuio-cloud:ROOT:references/quality-requirements/performance/ns-quota.adoc[Namespace quota per Zone]
 *** xref:appuio-cloud:ROOT:references/quality-requirements/performance/resource-quota.adoc[Resource limit per Namespace]
+*** xref:appuio-cloud:ROOT:references/quality-requirements/performance/portal-api-data-freshness.adoc[Portal API Data Freshness]
 
 ** Reliability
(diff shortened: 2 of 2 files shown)


docs/modules/ROOT/pages/references/quality-requirements/performance/portal-api-data-freshness.adoc
```diff
@@ -0,0 +1,27 @@
...
+{product} Portal API client
+
+Artifact::
+Portal API client and Portal API
+
+Response::
+Portal API returns the requested data
+
+Response measure::
+The read data is at most 10 minutes older than the original data in the backend system (source-of-truth)
```
corvus-ch, line 19: I do not see a need for such a short duration. Depending on the data, I might also be OK with data that is 24 hours old.

Consider the payment status of an invoice that is supposed to be payed through bank transfer. The Banks most likley will provide updated data once every day. So what is the deal with aiming for a 10 minutes freschness when there is nothing to refresh?

There might be data where freshness makes a difference, but I disagree with phrashing it in such a general way.

corvus-ch (changes requested):

ccremer (changes requested):
This requirement feels "wrong". I imagine this is about usability: The user requesting the data shall be informed about the "data" (We should probably clarify what kind of data we are talking about here, e.g. concrete invoices). "Data" is very generic.

Regardless of that, I'm not sure if APPUiO cloud should specify any "freshness" requirement because it's simply not its responsibility:
* If the Portal is REST-like synchronous and directly accessing the third-party system: Data retrieval can fail, thus an error message shall be displayed
* If the Portal is storing data CRD-like, then we simply show a "last updated time" or so. Because the ERP-specific operator might fail. Otherwise with a fixed "max freshness" every Operator will need to provision a HA ERP system which is probably not intended here in this eventual-consistency context. Do you understand what I mean :smile: ?

docs/modules/ROOT/pages/references/quality-requirements/performance/portal-api-data-freshness.adoc
tobru: This requirement talks about the freshness between the backend system and the API, not what and when the backend system retrieves third-party external data.

tobru:
I'm also not sure if we really need that requirement, I felt that's needed based on our discussion we had. I'm OK with not agreeing on this requirement and closing it for the time being.

ccremer:
Well, the "freshness" contradicts a bit #37 in some regards.
However, if we turn this idea into the usability one that i mentioned before, I'm fine with rewording in a way that it becomes relevant as a NEW requirement out of the decision in #35  (e.g. error message vs "last updated timestamp")

So, we can either close it or keep it open/draft until #35 is definitive :shrug:

tobru:
Let's keep it open a bit until #35 has settled down and decide afterwards.

docs/modules/ROOT/pages/references/quality-requirements/performance/portal-api-data-freshness.adoc
corvus-ch: Let me phrase it differently. With that requirement, an implementation has to run at least every 10 minutes. That is a waste of resources when it is known that there won't be any changes because the backend system gets updated only once a day.

corvus-ch: No that I have read the wording more carefully, it is about the time delta a change in the backend gets reflected in the API. But then again, why? It could be that we have cases where a short delta is important. For the examples I have in mind, it does not matter and would even make things more complicated.

ccremer: @tobru  and I did briefly discuss this in a call today about that. We realized that we need to split this requirement into 2. Since "data" is too generic we misunderstood its impact.
- One requirement deals with the entities like User, Organization or Teams that gets synced to _all zones_. There we need faster sync intervals.
- For invoices and other externally-managed entities, which are only in _appuio global_ we can loosen the restriction to $tbd hours or so.

corvus-ch (changes requested):
I flag this as "needs changes". Please request anotehr review once I shall have another look at it.

tobru:
After some time has passed and new insights have been made, I propose to close this PR. I don't see a need for it at the moment anymore.

PR #37, opened 2021-09-27
Portal API data recoverability requirement

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
+++ b/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
@@ -1,0 +1,24 @@
+= Portal API Data Recoverability
+
+Source::
+Portal API
+
+Stimulus::
+Portal API data is lost or is being freshly initialized
+
+Environment::
+Portal API
+
+Artifact::
+Portal API
+
+Response::
+Portal API retrieves all data from backend systems and makes the data available again
+
+Response measure::
+All data is available again as it was before the data loss
+
+Rationale::
+The Portal API doesn't store any relevant data.
+All relevant data is available and stored in third-party backend systems.
+The source of truth of the data is outside of the Portal API.
--- a/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
+++ b/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
@@ -1,24 +1,21 @@
-= Portal API Data Recoverability
+= Portal API Ephemeral Data
 
 Source::
 Portal API
 
 Stimulus::
-Portal API data is lost or is being freshly initialized
+Data is stored
 
 Environment::
 Portal API
 
 Artifact::
-Portal API
+Portal API and external third-party systems
 
 Response::
-Portal API retrieves all data from backend systems and makes the data available again
-
-Response measure::
-All data is available again as it was before the data loss
+Portal API stores the data in an ephemeral storage while data persistency is granted by the third-party backend system
 
 Rationale::
-The Portal API doesn't store any relevant data.
+The Portal API doesn't store any relevant data persistent.
 All relevant data is available and stored in third-party backend systems.
 The source of truth of the data is outside of the Portal API.
--- a/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
+++ b/docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
@@ -1,21 +1,26 @@
-= Portal API Ephemeral Data
+= Data compatibility with external components
 
 Source::
-Portal API
+{product} administrator
 
 Stimulus::
-Data is stored
+Reinstall APPUiO Global from scratch
 
 Environment::
-Portal API
+{zone}
 
 Artifact::
-Portal API and external third-party systems
+Portal API, external third-party component
 
 Response::
-Portal API stores the data in an ephemeral storage while data persistency is granted by the third-party backend system
+Entities that have their lifecycle outside of {product} have their availability restored
+
+Response measure::
+Following entities are recovered at latest after 48 hours since reinstallation:
++
+* Invoices
 
 Rationale::
-The Portal API doesn't store any relevant data persistent.
-All relevant data is available and stored in third-party backend systems.
-The source of truth of the data is outside of the Portal API.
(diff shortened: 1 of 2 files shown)


corvus-ch (changes requested):
This is lacking context. I expect that someone not involved in our discussion, will not understand this.

I would mor aim into the direction to have the API as an ephmeral system or if it uses a persistency, that the persistet data can be recovered from the origin system at any time.

tobru:
In which category of ISO would this requirement fall into?

ccremer (changes requested):
I agree with @corvus-ch 
This feels like the requirement was written afterwards to justify a choice in technology. However, it should rather be the other way round: the requirement is setting the playground for the choice of technology that fulfills the requirement.

I would also rather address the persistence aspect: APPUiO Cloud is not responsible for the lifecycle of data where such data is handled externally. In this context, it actually makes sense to define what kind of data we are talking about here (for example, metrics is still part of APPUiO, right?). That would be currently only invoices, right?

tobru:
I've reworded the requirement without moving the category yet, what category should I choose?

> In this context, it actually makes sense to define what kind of data we are talking about here (for example, metrics is still part of APPUiO, right?). That would be currently only invoices, right?

Not sure about that, I wouldn't list the data involved. Metrics for example are also stored in an external system.

ccremer:
> In which category of ISO would this requirement fall into?

Thinking of it, it's not that easy :smile: 
There are several:

* Reliability - Fault tolerance: "Degree to which a system, product or component operates as intended despite the presence of hardware or software faults." -> we want to highlight the absence of persistent storage, ergo be independent of storage failures, making the application stateless. (yes, CRDs or other cachings can be restored from scratch)
* Functional suitability - Functional appropriateness: "Degree to which the functions facilitate the accomplishment of specified tasks and objectives." -> with the lack of persistence we delegate the task of handling state to the external system
* Compatibility - Interoperability: "Degree to which two or more systems, products or components can exchange information and use the information that has been exchanged." -> focusing more on the data exchange without a storage layer between.

ccremer:
> Not sure about that, I wouldn't list the data involved. Metrics for example are also stored in an external system.

That "external system" is still part of APPUiO Cloud, right? We don't make this pluggable, but the same system is part of every APPUiO Cloud instance. In that case, it's an internal component not intended to be replaceable. For me, "external" in this context means not part of the system context (https://kb.vshn.ch/appuio-cloud/explanation/system/context.html).

The problem is, this requirement should be somewhat testable by a person (e.g. product owner). If it's not defined what kind of data we want to address here, how can this requirement ever be verified?

tobru:
> The problem is, this requirement should be somewhat testable by a person (e.g. product owner). If it's not defined what kind of data we want to address here, how can this requirement ever be verified?

Can't we find a generic description for that? "Data outside APPUiO Cloud system context"?

I'd vote fill this under "Compatibility - Interoperability".

docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
```diff
@@ -0,0 +1,21 @@
...
+Artifact::
+Portal API and external third-party systems
+
+Response::
+Portal API stores the data in an ephemeral storage while data persistency is granted by the third-party backend system
+
+Rationale::
+The Portal API doesn't store any relevant data persistent.
+All relevant data is available and stored in third-party backend systems.
+The source of truth of the data is outside of the Portal API.
```
ccremer, line 21:
```suggestion
= Data compatibility with external components

Source::
{product} administrator

Stimulus::
Reinstall APPUiO Global from scratch

Environment::
{zone}

Artifact::
Portal API, external third-party component

Response::
Entities that have their lifecycle outside of {product} have their availability restored

Response measure::
Following entities are recovered at latest after 48 hours since reinstallation:
+
* Invoices

Rationale::
{product} doesn't store any relevant third-party entities.
Although some form of caching is permissible, the cache can be rebuilt entirely in the given timeframe by reintegrating the third-party component.
The source of truth of the mentioned entities is outside of {product}.
```

ccremer (commented):
I tried to come up with a different phrasing.
WDYT?

docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
```diff
@@ -0,0 +1,26 @@
...
+{zone}
+
+Artifact::
+Portal API, external third-party component
+
+Response::
+Entities that have their lifecycle outside of {product} have their availability restored
+
+Response measure::
+Following entities are recovered at latest after 48 hours since reinstallation:
```
tobru, line 19:
```suggestion
Following entities are recovered at latest after 8 hours since reinstallation:
```

I feel 48 hours is way too long, it should be possible to restore much faster.

docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
```diff
@@ -0,0 +1,26 @@
...
++
+* Invoices
```
tobru, line 21:
```suggestion
* Invoices
* List of APPUiO Zones
* Organizations
* Teams
* Users
```

I really don't like to have this list here as I'm pretty sure that this list falls behind very fast. I'd rather find a generic description of this list, like "APPUiO Cloud Core Entities".

"The {product} core entities are recovered at latest after X hours since reinstallation"

We can then have another page describing the core entities which we can amend as the architecture grows without having to change a requirement.

ccremer: I think we may have a misunderstanding. Organizations, teams and users are stored in Keycloak and thus are not external data. To my understanding, this requirement doesn't need to cover those. Otherwise you'd be addressing disaster recovery processes, right? We haven't talked about that yet.
This requirement is about restoring data and processes where APPUiO Cloud has no data authority thereof.

docs/modules/ROOT/pages/references/quality-requirements/reliability/portal-api-data-recoverability.adoc
```diff
@@ -0,0 +1,26 @@
+= Data compatibility with external components
+
+Source::
+{product} administrator
+
+Stimulus::
+Reinstall APPUiO Global from scratch
```
ccremer, line 7: I think I know where the misunderstanding could come from. "Reinstall APPUiO Global from scratch" is too broad...
```suggestion
Reinstall APPUiO Global third-party component from scratch
```

PR #39, opened 2021-09-30
Introduce term 'Adapter' and 'Control API'

ccremer:
This should help using common language in discussions and documentation

## Checklist


docs/modules/ROOT/pages/references/glossary.adoc
```diff
@@ -15,18 +15,40 @@ It's the entity that owns the associated resources and the business connection.
...
+Control API::
+An API endpoint that exposes certain {product} features to users.
+Specifically, this consists of:
+* API that users write to with various clients (for example frontend or CLI) and thus configure their organizations
+* API that users read with various clients that is usually the result of {product} components (for example read-only access to invoices)
+
+Plugin API::
+A subset of Control API that is specifically targeted for Plugins.
+This specification may not only be technical, but may also contain a "behavioral contract" that is in documentation only.
+It basically defines how Plugins should behave and interact with {product}.
```
ccremer, line 39: I'm not quite sure about the `a "behavioral contract" that is in documentation only`. With this sentence, I want to point out to the metering system idea: The plugin API in this case defines the contract how the metrics database is to be accessed, e.g. database and data scheme. In that case I consider the data scheme from $database to be the "API" that the plugin has to understand in order to make sense of the data.

I would be open to exclude this type of "contract" from this term and/or define another term for that though.

tobru (approved):
I like this definition and I feel that's safe enough to try.

corvus-ch (approved):
"Adapter" might be the better term for "Plugin". However, I can not argue why.

tobru:
> "Adapter" might be the better term for "Plugin". However, I can not argue why.

Oh, yes, I'd favor "Adapter" over "Plugin" much more. Wdyt @ccremer ?

ccremer:
Both are fine by me.
"Plugin" in my understanding refers to the ability to plug in to an existing setup defined by an API.
"Adapter" refers to the ability to, well, adapt expected behavior against another system.

So, a $ERP operator would be both plugin and adapter at the same time, so :shrug: but let's agree on one term to speak the same language.

PR #49, opened 2021-10-15
Control API Design Documentation

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
@@ -1,0 +1,74 @@
+= APPUiO Control API Design
+
+This page describes the overall APPUiO Control API ("API" in the rest of this page) design.
+It adheres to the decision taken in xref:explanation/decisions/control-api.adoc[APPUiO Control API].
+The design follows the features asked in xref:explanation/system/details-ui.adoc[System idea: Graphical user interface].
+
+image:reference/control-api-design.drawio.svg[]
+
+== Cross-Zone Objects
+
+The cross-zone objects are stored in a control namespace called `appuio-cloud-control`.
+These are the objects which are global to the {product} instance.
+RBAC rules do protect these objects.
+
+=== `Organization`
+
+This is the main object.
+It's job is to manage the organization.
+
+* Manage the lifecycle and access rules of the Organization specific Namespace
+* Manage the organization in {idp}
+
+[source,yaml]
+----
+apiVersion: appuio.io/v1
+kind: Organization
+metadata:
+  name: acme-corp
+  namespace: appuio-cloud-control
+spec:
+  displayName: Acme Corp.
+status:
+  namespace: org-acme-corp
+  keycloak-group: acem-corp
+----
+
+The adapter to this object is core of {product}.
+
+=== `AppuioZone`
+
+This is a list of all {zone}s.
+
+[source,yaml]
+----
+apiVersion: appuio.io/v1
+kind: AppuioZone
+metadata:
+  name: cloudscale-lpg-0
+  namespace: appuio-cloud-control
+spec:
+  displayName: cloudscale.ch LPG 0
+  features:
+    openshift-version: "4.8"
+    kubernetes-version: "1.21"
+    sdn: OVN-Kubernetes
+  console-url: https://console.cloudscale-lpg-0.appuio.cloud/
+  kubernetes-api-url: https://api.cloudscale-lpg-0.appuio.cloud:6443/
+status:
+  status-message: running
+----
+
+The adapter to this object is specific to the reseller of {product}.
+
+== Organization Specific Objects
+
+Each organization has a dedicated namespace which contains the organization-level objects.
+These namespaces are named like `org-$orgname`.
+
+RBAC rules are deployed on these namespaces to protect access to the objects.
+
+Example:
+
+* Invoices
+* Teams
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+kind: Invitation
+metadata:
+  name: invite-1
+  namespace: org-acme-corp
+spec:
+  invitationcode: ju1aefaeKooz <1>
+  role: org-viewer <2>
+----
+<1> Generated invitation code
+<2> Role which this invitation assigns
```
ccremer, line 132: Who or what creates this invitation? Is it the org admin? Would we have to enforce rules like "min password length = 30 chars, uppercase, numbers etc"?
Regardless of that, with sharing some sort of code or token we would open up guessing/brute force attack surface.

Also, any other member of this org could share the code with someone else, that someone else could then hijack the invitation. could be even just an accidental leak from the org admin itself.

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+spec:
+  invitationcode: ju1aefaeKooz
+status:
+  blubb: bla
+----
+. The invitation controller searches through all namespaces to find an `Invitation` object where `spec.invitationcode` is the same
+.. A clusterrole and a clusterrolebinding is generated to allow access to this `InvitationAccept` object, granting access to the creating user
+.. The user is added to the matching rolebinding (`spec.role` in `Invitation`)
+.. The `Invitation` and the `InvitationAccept` object gets deleted
+.. The clusterrole and a clusterrolebinding is deleted
```
ccremer, line 150: Alternative idea:
1. Org admin creates invitation with a spec like this:
```yaml
apiVersion: appuio.io/v1
kind: Invitation
metadata:
  name: invite-joe
  namespace: org-acme-corp
spec:
  user: joe.mustermann
  role: org-viewer
status:
  accepted: false
```
2. Invitation controller creates a Role/Rolebinding that allows `joe.mustermann` to view and edit `invite-joe`
3. `joe.mustermann` edits the invite, sets `accepted: true` (alternatively, accept invitation with dedicated `InvitationAccept` resource that refers which invitation, e.g. `spec.invitationRef=invite-joe`)
4. Kyverno policy verifies that only the user in the spec can edit this invitation and set `accepted: true`
5. After accepting the edit, `Invitation` and Rolebinding gets deleted

This would leverage the Kubernetes/Openshift Authentication so that only the targeted user can accept the invitation.

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+  name: cloudscale-lpg-0
+spec:
+  displayName: cloudscale.ch LPG 0
+status:
+  features:
+    openshift-version: "4.8"
+    kubernetes-version: "1.21"
+    sdn: OVN-Kubernetes
+  console-url: https://console.cloudscale-lpg-0.appuio.cloud/
+  kubernetes-api-url: https://api.cloudscale-lpg-0.appuio.cloud:6443/
```
ccremer, line 170: Why is the display name a `.spec` but all the other features are `.status`?

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
+= APPUiO Control API Design
+
+This page describes the overall APPUiO Control API ("API" in the rest of this page) design.
+It adheres to the decision taken in xref:explanation/decisions/control-api.adoc[APPUiO Control API].
+The design follows the features asked in xref:explanation/system/details-ui.adoc[System idea: Graphical user interface].
+
+image:reference/control-api-design.drawio.svg[]
+
```
ccremer, line 8: I actually don't understand what this diagram is trying to visualize :see_no_evil:  but let's discuss that separately.

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+The benefit of providing these resources instead of only using CRDs is that we can calculate access permissions dynamically for every request.
+The same concept is also used by OpenShift with its `Project` resource which represents RBAC filter `Namespaces` (see https://github.com/openshift/kube-projects[kube-projects]).
+
+=== Authentication and Authorization
+
+Authentication against the API server is done by the {idp} by leveraging OIDC.
+This allows to transport authorized information about the user in a JWT token.
+
+For authorization, standard https://kubernetes.io/docs/reference/access-authn-authz/rbac/[Kubernetes RBAC] is being used.
+Kyverno policies can be used to implement enhanced policies, for example the number of resources of a specified kind a user is allowed to create.
```
ccremer, line 36: What this page doesn't cover is how the first access after registration works. Openshift first has to know the user before being able to authorize users or before Rolebindings take effect. Do newly registered users have to first login to the console of $someCluster (so that the user object gets created) before being able to use the API (and by extension, the Web UI)?

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+These are the resources which are global to the {product} instance.
+They are available in the global scope of the Kubernetes API server.
+
+=== `Organization` (Virtual)
+
+Organizations are the container for all resources which belong to an organization.
+It is represented by a Kubernetes Namespace with specific labels and annotations.
+This allows to use standard Kubernetes RBAC rules for organization-specific resources.
+
+To circumvent the current limitation of Kubernetes to show filtered lists of cluster-scoped resources based on access rights, the virtual object `Organization` represents an organization.
```
ccremer, line 69: Can you rephrase that? I don't understand how `the virtual object Organization represents an organization.` circumvents RBAC limitations?

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+* See details of zone I have access to
+
+== Global Resources
+
+These are the resources which are global to the {product} instance.
+They are available in the global scope of the Kubernetes API server.
+
+=== `Organization` (Virtual)
+
+Organizations are the container for all resources which belong to an organization.
```
ccremer, line 65: otherwise this sentence makes little sense.
```suggestion
`Organization` objects are the container for all resources which belong to an organization.
```

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+kind: Namespace
+metadata:
+  name: org-acme-corp <1>
+  labels:
+    appuio.io/resource-type: organization <2>
+    organization.appuio.io/displayName: Acme Corp. <3>
+----
+<1> Prefixed (`org`) name to circumvent possible name collision (e.g. if a company is called "kube-system")
+<2> Identify resource type, used by the API server to filter for namespace representing organizations
+<3> Reflected in the organization object as `spec.displayName`
```
ccremer, line 95: This part is missing a bit of context? (please rephrase if I understood it wrong)
```suggestion
The dynamic filter works through the list of namespaces which represent an organization and checks if the subject has access to the namespace.

Consider a user-generated `Organization`:
[source,yaml]
----
apiVersion: appuio.io/v1
kind: Organization
metadata:
  name: acme-corp
spec:
  displayName: Acme Corp.
----

The API server would then create a `Namespace` with following spec:
[source,yaml]
----
apiVersion: v1
kind: Namespace
metadata:
  name: org-acme-corp <1>
  labels:
    appuio.io/resource-type: organization <2>
    organization.appuio.io/displayName: Acme Corp. <3>
----
<1> Prefixed (`org`) name to circumvent possible name collision (e.g. if a company is called "kube-system")
<2> Identify resource type, used by the API server to filter for namespace representing organizations
<3> Reflected in the organization object as `spec.displayName`
```

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
tobru: The API doesn't have to run on OpenShift. I expect this API to run on a lightweight Kubernetes like vcluster.

docs/modules/ROOT/pages/references/architecture/control-api-design.adoc
tobru: With that the org admin would already have to know the details of the user to invite, and it could only target 1 user. I guess we need to clearly define how the invitation feature should work in general (use-case based) before we can create a design. And it might make sense to have that design in a separate page, focusing on it.

PR #60, opened 2021-11-02
Rewrite resource-quota requirement

simu:
* Remove specific numbers
* Add distinction between `NotTerminating` and `Terminating` pods

## Checklist


ccremer (changes requested):
Without the numbers this requirement will be hard to measure and verify from a PO's perspective.
I do see the problem of maintaining duplicated numbers both here and in the implementation though...
Please link at least the defaults where one can follow up on the numbers (https://github.com/appuio/component-appuio-cloud/blob/master/class/defaults.yml)

tobru (approved):

ccremer (approved):

PR #61, opened 2021-11-03
Correctly specify that we only consider CPU/memory limits for Terminating Pods

simu:
See also https://github.com/appuio/component-appuio-cloud/pull/32

## Checklist


simu:
This matches the setup on APPUiO Public. I didn't label the PR because we already have #60 for the changelog

tobru (approved):

PR #65, opened 2021-11-05
Decision about Unit Prefixes

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanation/decisions/unit-prefixes.adoc
+++ b/docs/modules/ROOT/pages/explanation/decisions/unit-prefixes.adoc
@@ -1,0 +1,22 @@
+= Unit prefixes
+
+== Problem
+
+On a low-level we work with bytes.
+Representing bytes to a human isn't practical, therefor a unit prefix has to be used to make it better understandable.
+
+== Proposals
+
+There are two well-known unit prefixes we could use:
+
+. https://en.wikipedia.org/wiki/Binary_prefix[binary prefixes]
+. https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes[SI prefixes]
+
+== Decision
+
+We use binary prefixes.
+
+== Rationale
+
+* In IT binary prefixes are representing the correct data.
+* OpenShift uses binary prefixes by default in their Web console.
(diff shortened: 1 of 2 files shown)


tobru:
Relates to https://github.com/appuio/appuio-cloud-docs/pull/49

docs/modules/ROOT/pages/explanation/decisions/unit-prefixes.adoc
```diff
@@ -0,0 +1,22 @@
+= Unit prefixes
+
+== Problem
+
+On a low-level we work with bytes.
+Representing bytes to a human isn't practical, therefor a unit prefix has to be used to make it better understandable.
+
```
ccremer, line 7: Maybe extend the problem desc that this decision would also apply to product specifications and marketing and maybe other places?

ccremer (commented):
I do not see any actual impact on the architecture. I merely see this decision as a form of definition that affects product documentation and maybe API specifications and maybe marketing.

Why not simply update https://products.docs.vshn.ch/products/appuio/cloud/price_model.html?

Edit: https://github.com/appuio/appuio-cloud-docs/pull/49 should be enough IMO

docs/modules/ROOT/pages/explanation/decisions/unit-prefixes.adoc
tobru: Done.

tobru:
The impact is given, as we have several places where we need to present "bytes" to a human: The Portal, Invoices, ... And as it already caused confusion because this wasn't documented, I think it is important that we write this down.

corvus-ch (approved):

ccremer:
> The impact is given, as we have several places where we need to present "bytes" to a human: The Portal, Invoices, ... And as it already caused confusion because this wasn't documented, I think it is important that we write this down.

Yes, there's an impact in various places like UI and invoices and we should write it down. It just doesn't have an impact _on the architecture_. I'ts more like a business decision. Displaying GiB instead of GB doesn't fundamentally change how core APPUiO Cloud works.

I propose to add a few sentences [here](https://kb.vshn.ch/appuio-cloud/references/architecture/metrics-of-interest.html#_metric_memory_usage) or  [here](https://kb.vshn.ch/appuio-cloud/references/architecture/metering.html)

tobru:
As I don't see another good place to document it, I go forward and merge it. No harm will be done by that if we have to move it to a different place.

PR #38, opened 2021-09-27
Replaceability of external components

tobru:
## Checklist

- [X] Try to isolate changes into separate PRs (to build a better changelog).
- [X] Categorize the PR by setting a good title and adding one of the labels:
      `change`, `decision`, `requirement/quality`, `requirement/functional`, `dependency`
      as they show up in the changelog

PR #68, opened 2021-11-11
Add how-to page to override namespace quota

ccremer:
Updates user documentation with the introduction of https://github.com/appuio/component-appuio-cloud/pull/38

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
+++ b/docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
@@ -1,0 +1,48 @@
+= Override Namespace Quota
+
+This guide describes how a cluster admin can override the namespace/project quota for an organization in a single {zone}.
+
+== Prerequisites
+
+* Administrator access to the {zone}
+
+== Steps
+
+=== Method 1: Configure in cluster catalog
+
+This method uses the https://hub.syn.tools/appuio-cloud/references/parameters.html#_namespacequotaoverrides[component parameters] to generate and maintain the `ConfigMaps` for the overrides.
+
+. Configure component per {zone}:
++
+[source,yaml]
+----
+parameters:
+  appuio_cloud:
+    namespaceQuotaOverrides:
+      example-company: 50 <1>
+----
+<1> The organization `example-company` can now have 50 namespaces
+
+. Compile and push the cluster catalog
+
+NOTE: ArgoCD will overwrite any existing `ConfigMaps` that may have been created manually.
+
+=== Method 2: Create ConfigMap manually
+
+The override is active as soon as you create a `ConfigMap` for the organization.
+
+[source,yaml]
+----
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: appuio-cloud <1>
+  name: override-example-company <2>
+data:
+  namespaceQuota: '50' <3>
+----
+<1> The namespace in which the component is installed.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+== Prerequisites
+
+* Administrator access to the {zone}
+
+== Steps
+
+=== Method 1: Create ConfigMap manually
+
+The override is active as soon as you create a config map for the organization, but only for the next namespaces to be created.
```
simu, line 13:
```suggestion
The override is active as soon as you create a `ConfigMap` for the organization.
```

I'm not sure the second part of the sentence is required at all, it should be obvious to a cluster admin that such a change won't retroactively allow namespaces to have been created in the past (unless I'm misunderstanding what you wanted to say with the second part).

docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
```diff
@@ -0,0 +1,46 @@
...
+  namespace: appuio-cloud <1>
+  name: override-example-company <2>
+data:
+  namespaceQuota: '50' <3>
+----
+<1> The namespace in which the component is installed.
+<2> The name follows the pattern `override-?` where `?` is the organization name as defined in `appuio.io/organization` annotation.
+<3> Set the new quota (as a string).
+
+TIP: `oc -n <namespace> create cm override-<org> --from-literal=namespaceQuota=10`
```
simu, line 29:
```suggestion
TIP: You can use `oc -n <namespace> create cm override-<org> --from-literal=namespaceQuota=10` to create an appropriate `ConfigMap` with a single command.
```

docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+=== Method 2: Configure in cluster catalog
```
simu, line 31: It would be nice to explicitly document what happens when I configure a ConfigMap via the cluster catalog which already exists on the system. I assume that ArgoCD will just adopt and update the existing ConfigMap in that case.

Also, I think we should promote doing such configurations via the cluster catalog as the first method.

simu (changes requested):

docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
ccremer: First part: I agree.
Second: In https://github.com/appuio/component-appuio-cloud/pull/38 we discussed that neither approach is actually preferred. Configuring it via Commodore is a method that should be replaced later with a better APPUiO-native mechanism

simu (approved):

docs/modules/ROOT/pages/how-to/day2ops/override-namespace-quota.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+== Prerequisites
+
+* Administrator access to the {zone}
+
+== Steps
+
+=== Method 1: Configure in cluster catalog
+
+This method uses the component parameters to generate and maintain the `ConfigMaps` for the overrides.
```
tobru, line 13: Maybe link to the component (docs) for reference?

tobru (approved):

bastjan (approved):

PR #69, opened 2021-11-12
Document label/annotation allowlist

bastjan:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/namespace-ownership.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/namespace-ownership.adoc
@@ -14,6 +14,13 @@ metadata:
 EOF
 ```
 
+== Labels and Annotations
+
+Kubernetes, various controllers, and OpenShift use labels and annotations, such as `scheduler.alpha.kubernetes.io/node-selector`, on namespaces to manage internal state and some advanced config.
+We don't necessarily know the labels and annotations in advance. This makes denylists hard to implement.
+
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/references/architecture/namespace-ownership.adoc
```diff
@@ -14,6 +14,13 @@ metadata:
 EOF
 ```
 
+== Labels and Annotations
+
+Kubernetes, various controllers, and OpenShift use labels and annotations, such as `scheduler.alpha.kubernetes.io/node-selector`, on namespaces to manage internal state and some advanced config.
+We don't necessarily know the labels and annotations in advance. This makes denylists hard to implement.
+
+Allowed labels are thus explicitly tracked in an allowlist. `custom.appuio.io/*` was chosen as a prefix for user defined labels.
```
ccremer, line 22:
```suggestion
Kubernetes, various controllers, and OpenShift use labels and annotations, such as `scheduler.alpha.kubernetes.io/node-selector`, on namespaces to manage internal state and some advanced config.
Such logic can have undesired effect over namespaces on a {zone}, thus they have to be managed using allowlists and denylists.
Denylists are hard to implement since the full set of undesired labels is not known in advance and they can change as OpenShift and Kubernetes evolve.

Allowed labels are thus explicitly tracked in an allowlist.
In order to enable users to add labels to namespaces, the prefix `custom.appuio.io/*` was chosen for user defined labels.
```

ccremer (approved):

PR #72, opened 2021-11-24
Add documentation regarding metering setup on a new zone

simu:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/how-to/zone-setup.adoc
+++ b/docs/modules/ROOT/pages/how-to/zone-setup.adoc
@@ -139,43 +139,46 @@ parameters:
         remoteWrite:
           - name: appuio-metering
             url: http://thanos-receive.appuio-metering-thanos-receiver.svc:19291/api/v1/receive
+            headers:
+              THANOS-TENANT: '${cluster:name}' <1>
             writeRelabelConfigs:
               - sourceLabels: ['__name__']
-                regex: '(container_memory_usage_bytes|kube_pod_container_resource_requests|kube_persistentvolumeclaim_resource_requests_storage_bytes)' <1>
+                regex: '(container_memory_usage_bytes|kube_pod_container_resource_requests|kube_persistentvolumeclaim_resource_requests_storage_bytes)' <2>
                 action: keep
 
   appuio_metering_thanos_receiver:
     namespace: appuio-metering-thanos-receiver
-    cluster_kubernetes_version: '${dynamic_facts:kubernetesVersion:major}.${dynamic_facts:kubernetesVersion:minor}' <2>
+    cluster_kubernetes_version: '${dynamic_facts:kubernetesVersion:major}.${dynamic_facts:kubernetesVersion:minor}' <3>
     commonConfig:
       volumeClaimTemplate:
         spec:
-          storageClassName: ssd <3>
+          storageClassName: ssd <4>
       # Override default security context for all components
       securityContext: {}
     objectStorageConfig:
       type: S3
       config:
-        bucket: '${cluster:name}-appuio-metering' <4>
-        endpoint: objects.lpg.cloudscale.ch <5>
-        region: lpg <5>
-        access_key: '?{vaultkv:${cluster:tenant}/${cluster:name}/cloudscale/s3_access_key}' <6>
-        secret_key: '?{vaultkv:${cluster:tenant}/${cluster:name}/cloudscale/s3_secret_key}' <6>
+        bucket: '${cluster:name}-appuio-metering' <5>
+        endpoint: objects.lpg.cloudscale.ch <6>
+        region: lpg <6>
+        access_key: '?{vaultkv:${cluster:tenant}/${cluster:name}/cloudscale/s3_access_key}' <7>
+        secret_key: '?{vaultkv:${cluster:tenant}/${cluster:name}/cloudscale/s3_secret_key}' <7>
     query:
       enabled: false # we don't need the querier on the zones
     receive:
       enabled: true
       replicas: 2
 ----
-<1> The set of metrics to collect (as an RE2 regex)
-<2> The cluster's Kubernetes version.
+<1> Send extra `THANOS-TENANT` header so that Thanos Receiver sets a useful `tenant_id` label on the received metrics.
+<2> The set of metrics to collect (as an RE2 regex)
+<3> The cluster's Kubernetes version.
 The example makes use of the `kubernetesVersion` dynamic fact reported by Steward on the cluster.
-<3> The storage class to use for the PVCs for the Thanos Receiver.
+<4> The storage class to use for the PVCs for the Thanos Receiver.
 We highly recommend using a storage class suitable for database workloads.
-<4> The name of the bucket created in the previous step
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
 
 . Compile and push the cluster catalog
 . Wait for Argo CD to sync the config
+
+== Configure metering metrics collection
+
+We collect a small set of metrics which are used for metering from each zone.
```
ccremer, line 123:
```suggestion
We collect a small set of metrics which are used for metering from each {zone}.
```

docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
...
+This is done by deploying a https://thanos.io/tip/components/receive.md/[Thanos Receiver] on each zone and configure the zone's cluster monitoring to send selected metrics to the Thanos Receiver via Prometheus's remote write feature.
+
+. Create an S3 bucket which can be accessed using the same credentials as the cluster's registry bucket.
+Use name `${CLUSTER_ID}-appuio-metering` for the bucket.
```
ccremer, line 127: small name suggestion. `metering` is for me a device or component to measure something. However the metrics we put into the bucket are already measured :)
```suggestion
Use name `${CLUSTER_ID}-appuio-metrics` for the bucket.
```

docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
...
+    commonConfig:
+      volumeClaimTemplate:
+        spec:
+          storageClassName: ssd
+      # Override default security context for all components
+      securityContext: {}
+    objectStorageConfig:
+      type: S3
+      config:
+        bucket: c-cluster-id-1234-appuio-metering <3>
```
ccremer, line 158:
```suggestion
        bucket: c-cluster-id-1234-appuio-metrics <3>
```

ccremer (approved):

docs/modules/ROOT/pages/how-to/zone-setup.adoc
simu: I've used metering everywhere, since that's the term which is used in the stories.

ccremer: Well, we use both terms :shrug: 
See e.g. https://kb.vshn.ch/appuio-cloud/references/architecture/metering.html

But I don't think we should be using `metrics` and `metering` interchangeably (as synonyms). For me, metering is the whole process, whereas metrics are just the data part of metering.

@corvus-ch Any opinions?

corvus-ch: For me the term `metering` refers to metrics that are used for statistics and invoicing. `metrics` is more bound to metrics used for monitoring and alerting. That is just my take on it. Whatever naming we choose, we must be able to distinguish between those cases. @mhutter is working on a centralized monitoring that builds on the same tools. Those must remain separated setups and their naming should make it clear.

docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
...
+
+== Configure metering metrics collection
+
+We collect a small set of metrics which are used for metering from each zone.
+This is done by deploying a https://thanos.io/tip/components/receive.md/[Thanos Receiver] on each zone and configure the zone's cluster monitoring to send selected metrics to the Thanos Receiver via Prometheus's remote write feature.
+
+. Create an S3 bucket which can be accessed using the same credentials as the cluster's registry bucket.
+Use name `${CLUSTER_ID}-appuio-metering` for the bucket.
+
+. Add configuration
```
corvus-ch, line 129: Is there a reason to not add this to the zone configuration?

docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
...
+        remoteWrite:
+          - name: appuio-metering
+            url: http://thanos-receive.syn-thanos.svc:19291/api/v1/receive
+            writeRelabelConfigs:
+              - sourceLabels: ['__name__']
+                regex: '(container_memory_usage_bytes|kube_pod_container_resource_requests|kube_persistentvolumeclaim_resource_requests_storage_bytes)' <1>
+                action: keep
+
+  thanos:
+    cluster_kubernetes_version: '1.22' <2>
```
corvus-ch, line 148: Why not use dynamic facts?

corvus-ch (commented):
I am probably miss some context but this feels to add manual steps which could be covered through the zone configuration in our hierarchy.

As for the naming, `metering` seems to be the way to go.

docs/modules/ROOT/pages/how-to/zone-setup.adoc
ccremer: Hm, if metrics is only for monitoring and alerting, wouldn't that be inconsistent with https://kb.vshn.ch/appuio-cloud/references/architecture/metrics-of-interest.html and ff?
Maybe we need to better define those terms in a separate PR.
Let's go with the `appuio-metering` bucket name for now, it's not that much of an issue for me.

corvus-ch: I agree. And yes, this is bottering me too.

simu: I think we're confusing two levels of abstraction here. On one level of abstraction we have metrics which are relevant for invoicing (what I've been calling "metering metrics" or similar in commit messages and config comments), on the other more abstract level we have the "APPUiO Metering" component (for lack of a better name).

To make it clear to which component this bucket belongs, I've been using the `appuio-metering` suffix to indicate the system component which uses this bucket.

In any case, I agree that we should probably clearly define terms around metrics and metering, but that's clearly out of scope for this PR.

docs/modules/ROOT/pages/how-to/zone-setup.adoc
simu: It is part of the zone configuration, but this how-to is the generic one which doesn't assume that you've got access to the VSHN global defaults (as far as I understand).

docs/modules/ROOT/pages/how-to/zone-setup.adoc
```diff
@@ -117,3 +117,62 @@ parameters:
...
+----
+applications:
+  - thanos
+parameters:
+  openshift4_monitoring:
+    configs:
+      prometheusK8s:
+        remoteWrite:
+          - name: appuio-metering
+            url: http://thanos-receive.syn-thanos.svc:19291/api/v1/receive
```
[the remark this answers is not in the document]
simu: This will most likely be renamed as soon as component-thanos multi-instance support is release

corvus-ch (approved):

PR #77, opened 2021-12-02
Billing and Reporting Requirements

tobru:
## Checklist

- [ ] Link this PR to related issues if applicable.

### DIFF
--- a/docs/modules/ROOT/pages/references/functional-requirements/billing.adoc
+++ b/docs/modules/ROOT/pages/references/functional-requirements/billing.adoc
@@ -32,55 +32,48 @@ _Invoice period: 1.1. - 31.1. (44'640 minutes, timezone UTC)._
 |Discount
 |Total (Rounded)
 
-6+e|*Compute*
+6+|*Zone: cloudscale.ch LPG 2 - Namespace: my-app-prod*
 
-a|
-*Zone*: cloudscale.ch LPG 2 +
-*Namespace*: my-app-prod
+a|Compute
 
-Usage:
-
-* Min: 128 MiB
-* Max: 1034 MiB
-* Avg: 585.66 MiB
+Min: 128 MiB +
+Max: 1034 MiB +
+Avg: 585.66 MiB
 
 |7028
 |MiB
-|0.0000012432
+|CHF 0.0000012432
 |0 %
-|0.01
+>|CHF 0.01
 
-a|
-*Zone*: Exoscale GVA 1 +
-*Namespace*: my-app-test
+6+>|*Subtotal*: CHF 0.01
 
-Usage:
+6+|*Zone: Exoscale GVA 1 - Namespace: my-app-test*
 
-* Min: 302 MiB
-* Max: 3990 MiB
-* Avg: 1689.2 MiB
+a|Compute
+
+Min: 302 MiB +
+Max: 3990 MiB +
+Avg: 1689.2 MiB
 
 |16892
 |MiB
-|0.0000009043
+|CHF 0.0000009043
 |0 %
-|0.02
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/references/functional-requirements/billing.adoc
```diff
@@ -0,0 +1,172 @@
...
+|===
+|Description
+|Quantity
+|Unit of Measure
+|Unit Price
+|Discount
+|Total (Rounded)
+
+6+e|*Compute*
+
```
ccremer, line 36: I looked at the preview and can't help but wonder if a sorting by namespace makes more sense than sort by product category.

The current example lists all compute resources by zone and namespace.
Then it lists all storage resources by zone and namespace.
Then maybe LoadBalancer, or other categories, by zone and namespace and so on.

This makes it hard to follow what exactly a namespace is costing me, as I'd have to search and sum up resources from all over the place.

What if instead the line items are combined and summarized by namespace instead?

For example (just prosa ASCII art):
```
                                  Quantity | Unit  | ...
========================================================
zone: cloudscale.ch LPG 2
namespace: my-app-prod

Compute
--------------------------------------------------------
Usage (Min / Max / Avg): 
128 MiB / 1034 MiB / 585.66 MiB  |  7028   |  MiB  |  ...
--------------------------------------------------------

Persistent Storage (PVC)
--------------------------------------------------------
Usage (Avg): 
10 GiB                           |  100    |  GiB  |  ...
--------------------------------------------------------

Total for namespace
--------------------------------------------------------
                                                    0.02

========================================================
zone: Exoscale GVA 1
namespace: my-app-test

Compute
--------------------------------------------------------
Usage (Min / Max / Avg): 
...
```

Ofc, not all namespaces have storage. and some may have storage without compute. In those cases the listing for a category for each namespace just disappears. The "Total for namespace" then basically gives summarizes the cost for a namespace.

A decision-maker could then better decide where to optimize cost for applications. e.g it allows to say "this test app costs us x amount per month. Do we need it?"

corvus-ch: I second that idea.

corvus-ch (changes requested):
I would like to see the suggestion of @ccremer to be applied.

From my understanding of Odood, that could be doable. It would require the adapter to create and use catecories based on zone and namespaces.

tobru:
It absolutely makes much more sense to group by Zone/NS. I updated the example accordingly. Please have a look again @ccremer @corvus-ch 

Do we also need to talk about VAT here or is this something which has to be tackled by the ERP / Billing System?

docs/modules/ROOT/pages/references/functional-requirements/billing.adoc
```diff
@@ -0,0 +1,165 @@
...
+Account ID:: Hardcoded to 49.
+Currency ID:: Hardcoded to 6 (CHF).
+Invoice date:: Date of the invoice.
+Journal ID:: Hardcoded to 1.
+Name:: Name of the invoice.
+Payment term:: Hardcoded to 3.
+State:: Draft.
+User ID:: The ID of the user who created the invoice.
+
+=== Per Line Item
```
corvus-ch, line 101: I remember we talked about not to use the Odoo internal price table but use our own. For this I am missing a field to enter the chared amount.

tobru: Yes, that was missing. Added.

ccremer (approved):
Example invoice LGTM, though I can't judge if the fields are correct in terms of Odoo

corvus-ch (approved):

PR #80, opened 2021-12-23
Remove `org-` prefix for organization namespaces

glrf:
There does not seem to be consensus on what naming collisions this prefix
should prevent.

The prefix makes the API less intuitive. When working with organization
`foo`, I expect to be able to access its teams in namespace `foo` and
not namespace `org-foo`. It's even worse when there actually is a
namespace `foo` and I get a permission denied error.

## Checklist


corvus-ch (approved):
I agree with dropping the prefix. If I recall the discussion correctly, we were still not used to an Kubernetes API, where there is not a lot else going on. Having this encapsulated in a vCluster, we have less things to take care of.

ccremer (approved):

tobru (approved):

PR #82, opened 2021-12-23
Add Specification for Organization Access Control

glrf:
This is a proposal to manage access control for `Organization` through a dummy RBAC resource called `organizations.rbac.appuio.io`. In essence this works the same way as the implied permission check through `Namespaces` but it has the advantage of not having to give users access to the actual `Namespace` resource.

I had a hard time describing the architecture in text form. I'm open for any suggestions to make this easier to understand.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
@@ -77,7 +77,7 @@ The filter uses the following heuristic:
 We use standard Kubernetes role-based access control for `Organizations` with two distinct differences.
 
 . Access needs to be granted for `organizations` resources in API group `rbac.appuio.io` and not for resources in API group `organization.appuio.io`.
-Permissions can be configured through both `Roles` and `RoleBindings`, as well as `ClusterRoles` and `ClusterRoleBindings`-
+Permissions can be configured through both `Roles` and `RoleBindings`, as well as `ClusterRoles` and `ClusterRoleBindings`.
 Similarly to `Namespaces`, permissions configured by a `Role` in namespace `foo` only affects `Organization` `foo`.
 
 . For `list` and `watch` verbs the API server will only return resources that the user also has permission to `get`
@@ -87,7 +87,7 @@ Similarly to `Namespaces`, permissions configured by a `Role` in namespace `foo`
 ====
 The `rbac.appuio.io` API group allows us to delegate access control to the aggregate API server.
 With the new API group we can still use the powerful RBAC engine of Kubernetes, while bypassing the Kubernetes API server's access control.
-
+As a consequence, by introducing this logical RBAC group, no custom code needs to be written in order to implement the access control required for multi-tenancy.
 The Kubernetes API server will still perform classical access control for `organizations.organization.appuio.io` resources.
 In practice any user is allowed to perform any action on `organizations.organization.appuio.io` and access control is handled by the aggregate API server.
 ====
--- a/docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
@@ -15,7 +15,7 @@ It is assumed that the `Organization` resource is used for all operations, the r
 .Virtual resource
 [source,yaml]
 ----
-apiVersion: appuio.io/v1
+apiVersion: organization.appuio.io/v1
 kind: Organization
 metadata:
   name: acme-corp <1>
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
 * Label: `appuio.io/resource-type=organization`
-* Subject is bound to one of the defined `ClusterRole` resources.
 
-== RBAC and Cluster roles
+== Access Control
 
-These are `ClusterRole` resources which are bound to a subject by a namespaced `RoleBinding`:
+Access control for `Organizations` uses standard Kubernetes RBAC rules, but `Organizations` are represented by two different resources.
+
+. The `organizations.organizations.appuio.io` resource is the cluster-scoped resource we actually access.
```
ccremer, line 79: Who is `we` in this context? I think we should clearly state which actor is "accessing" them. we failed to previously document who is the main entity that manages or interacts with these resources.
```suggestion
. The `organizations.organizations.appuio.io` resource is the cluster-scoped resource any authenticated human user actually interacts with.
```

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
+The access control for this resource is done by the Kubernetes API server.
+RBAC works the same as for any other cluster-scoped resources and can be configured through `ClusterRoles` and `ClusterRoleBindings`.
+
+. The `organizations.rbac.appuio.io` resource is a custom namespace-scoped resource known only to the `Organization` API server, which is only used for access control.
+Being namespace-scoped, permissions can also be configured through `Roles` and `RoleBindings`.
+Similarly to `Namespaces`, permissions configured by a `Role` in namespace `foo` only affects `Organization` `foo`.
+
+A user can only perform an action on an `Organization`, if the action is allowed on both of these resources.
+The `list` and `watch` verbs are special cases and the API server will only return resources that the user also has permission to `get`
```
ccremer, line 88:
```suggestion
The `list` and `watch` verbs are special cases and the custom API server will only return resources that the user also has permission to `get`.
```

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
+
+In practice this means, we will allow any user to perform any action on `organizations.organizations.appuio.io` and manage access to `Organizations` through `organizations.rbac.appuio.io`.
```
ccremer, line 90:
```suggestion
In practice this means, any user is allowed to perform any action on `organizations.organizations.appuio.io` and manage access to `Organizations` through `organizations.rbac.appuio.io`.
```

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
+
+[NOTE]
+====
+This trick of using `organizations.rbac.appuio.io` allows us to handle and especially delegate access to `Organizations` on a namespace level.
+So we can allow users to configure access without having to expose `ClusterRoleBindings`.
```
ccremer, line 95:
```suggestion
Users can then configure access in their organizations without the permissions to maintain `ClusterRoleBindings`.
```

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
-== RBAC and Cluster roles
+== Access Control
 
-These are `ClusterRole` resources which are bound to a subject by a namespaced `RoleBinding`:
+Access control for `Organizations` uses standard Kubernetes RBAC rules, but `Organizations` are represented by two different resources.
+
+. The `organizations.organizations.appuio.io` resource is the cluster-scoped resource we actually access.
+The access control for this resource is done by the Kubernetes API server.
+RBAC works the same as for any other cluster-scoped resources and can be configured through `ClusterRoles` and `ClusterRoleBindings`.
+
```
ccremer, line 82: Reading through this, there is something missing here why we solve the problem we describe just below.

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,82 @@ metadata:
...
 `appuio-organization-admin`:: Admin (read / write) access to an organization
++
+[source,yaml]
+----
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: appuio-organization-admin
+rules:
+- apiGroups: ["rbac.appuio.io"]
```
ccremer, line 127: I'm not quite sure we actually need a custom group `rbac.appuio.io`. can the custom API server not just verify (in addition to standard Kubernetes RBAC) to see if there's additional permissions in a namespace-scoped rolebinding?

Maybe a namespace-scoped example is needed to better understand why this special group is needed? Because I'm guessing this is done to allow users of other organizations to access it?

ccremer (commented):
I still don't fully understand why the custom rbac group is needed though. Maybe you can give example specs that outlines the conflict you're trying to solve?

corvus-ch (approved):
I do like the idea to decouple access to the namespace from access to the organization. Given the remarks by @ccremer are resolved. This is looking good to me.

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
glrf: Theoretically yes we probably could add additional checks in our API server to see whether there is a rolebinding to explicitly allow accessing an organization. In practice I expect this to get hairy.

There is no way (that I found) to disable standard Kubernetes RBAC for organization. This means we will have to allow any action on organizations for everyone on a cluster scope. Otherwise our API server never gets the request. So we would only allow actions on organizations if the action is allowed on a cluster and namespace level. I see two issues with that:

* This is IMO not really more intuitive that adding another custom group and this will be annoying for cluster-admins as we do now not have any option to enable cluster wide access to organizations
* We need to implement parsing roles/rolebindings ourselves. Currently we can basically use `kubectl auth can-i` to check if a user is allowed to perform actions on a `organizations.rbac.appuio.io` resource. Re-implementing these security checks is not something I want to do as it's error prone and errors have security implications.

I will try to write something that better explains the use of this new group.

ccremer: Thanks for the clarification. I think especially the second bullet should actually make it into the page, as this is a quite important aspect IMO.

glrf:
I tried to restructure the access control explanation. I hope it's a little clearer now.

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,79 @@ metadata:
...
 * Label: `appuio.io/resource-type=organization`
-* Subject is bound to one of the defined `ClusterRole` resources.
 
-== RBAC and Cluster roles
+== Access Control
+
+We use standard Kubernetes role-based access control for `Organizations` with two distinct differences.
+
+. Access needs to be granted for `organizations` resources in API group `rbac.appuio.io` and not for resources in API group `organization.appuio.io`.
+Permissions can be configured through both `Roles` and `RoleBindings`, as well as `ClusterRoles` and `ClusterRoleBindings`-
```
ccremer, line 80: typo?
```suggestion
Permissions can be configured through both `Roles` and `RoleBindings`, as well as `ClusterRoles` and `ClusterRoleBindings`.
```

docs/modules/ROOT/pages/references/architecture/control-api-org.adoc
```diff
@@ -63,24 +63,79 @@ metadata:
...
+
+. For `list` and `watch` verbs the API server will only return resources that the user also has permission to `get`
+
 
-These are `ClusterRole` resources which are bound to a subject by a namespaced `RoleBinding`:
+[NOTE]
+====
+The `rbac.appuio.io` API group allows us to delegate access control to the aggregate API server.
+With the new API group we can still use the powerful RBAC engine of Kubernetes, while bypassing the Kubernetes API server's access control.
+
```
ccremer, line 90: Maybe also implementation consequences?
```suggestion
With the new API group we can still use the powerful RBAC engine of Kubernetes, while bypassing the Kubernetes API server's access control.
As a consequence, by introducing this logical RBAC group, no custom code needs to be written in order to implement the access control required for multi-tenancy.
```

ccremer (approved):

PR #92, opened 2022-01-06
Only expose the user preference in user spec

tobru:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/control-api-user.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/control-api-user.adoc
@@ -12,7 +12,8 @@ kind: User
 metadata:
   name: bec0d928-2ae2-4cec-94a0-5f72f12b8b39 <1>
 spec:
-  defaultOrganizationRef: acme-corp
+  preferences:
+    defaultOrganizationRef: acme-corp
 status: <2>
   displayName: Kate Demo
   username: kate.demo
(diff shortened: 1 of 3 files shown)


tobru:
@corvus-ch @ccremer As discussed: The `User` object only exposes the user preference. All other fields are read-only and exposed via the status sub-resource.

Some thoughts:
* We could move `spec.defaultOrganizationRef` to `spec.preferences.defaultOrganizationRef` to better distinguish from main user parameters, but I opted to keep it flat.
* I thought about renaming the object to `UserPreference`, but decided to keep it as `User` as it is basically representing the User and is used for other mechanisms which are referencing users.

docs/modules/ROOT/pages/references/architecture/control-api-user.adoc
```diff
@@ -12,8 +12,6 @@ kind: User
 metadata:
   name: bec0d928-2ae2-4cec-94a0-5f72f12b8b39 <1>
 spec:
-  displayName: Kate Demo
-  email: kate@demo.com
   defaultOrganizationRef: acme-corp
```
ccremer, line 15: If you would like to keep the resource named `User`, then I'd prefer a `spec.preferences.defaultOrganizationRef` key. Otherwise, it's not quite clear that this field is actually a preference and not a hard link/association to an organization.

tobru: fine by me

corvus-ch (approved):

ccremer (approved):

PR #93, opened 2022-01-06
Schema Clarifications

bastjan:
- `from/until` to `during` timerange. \
  From is an sql keyword and can't be used. Timerange allows nicer indexes, stores bounds.
- Clarify nullable fields
- Clarify inclusive/exclusive range bounds

## Checklist


corvus-ch (changes requested):
Please also add the definition of inclusive/exclusive for before/after to the text, as it is done for the unique constraints.

corvus-ch (approved):

PR #109, opened 2022-02-08
Add page referring to the appuio-cloud Commodore component Kyverno policy documentation

simu:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanation/kyverno-policies.adoc
+++ b/docs/modules/ROOT/pages/explanation/kyverno-policies.adoc
@@ -1,0 +1,7 @@
+= Kyverno policies used by {product}
+
+We use Kyverno as the xref:appuio-cloud:ROOT:explanation/decisions/kyverno-policy.adoc[policy engine] and xref:appuio-cloud:ROOT:explanation/decisions/kyverno-generator.adoc[resource generator] in {product}.
+
+We manage all Kyverno policies for {product} in the https://hub.syn.tools/appuio-cloud/index.html[appuio-cloud Commodore component].
+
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanation/kyverno-policies.adoc
```diff
@@ -0,0 +1,7 @@
+= Kyverno policies used by {product}
+
+As documented in xref:appuio-cloud:ROOT:explanation/decisions/kyverno-policy.adoc[this decision] and xref:appuio-cloud:ROOT:explanation/decisions/kyverno-generator.adoc[this decision], {product} uses Kyverno as its policy engine and resource generator.
```
corvus-ch, line 3: I was once looking into accessible web design. There I learned that links with texts like _here_, _this x_ etc. are an issue for people with disabilities. The suggestion was to replace them with something more telling. I realized that this then makes it also easier to read for people without disabilities. At least it was for me.

corvus-ch (approved):

docs/modules/ROOT/pages/explanation/kyverno-policies.adoc
simu: Yeah, I try to avoid using that style of links when possible, but I couldn't figure out a way to phrase it differently yesterday. I'll have a look again.

simu: I've rephrased this sentence, it should be much better now

corvus-ch: It most certainly is, thanks.

PR #113, opened 2022-03-09
Document migration from `id.appuio.cloud` to `id.vshn.net`

bastjan:
TODO: Decide if we should merge this as an example or close the PR after a successful migration.

Documents how to migrate APPUiO Cloud Clusters to `id.vshn.net`.

## Checklist


simu (approved):
Content looks good as far as I can tell. Structure could be improved a bit, see inline comments.

> TODO: Decide if we should merge this as an example or close the PR after a successful migration.

My vote is to keep the guide as an example for how to migrate APPUiO Cloud to a different Keycloak instance.

PR #122, opened 2022-05-02
Include fact properties in fact table

glrf:
Instead of adding another (sub-)fact table and linking the two fact tables, we integrate sub facts in the main fact table. This leaves us with a more standard star schema and simplifies data ingestion and querying.

The invoice generation differentiate facts from subfacts and find the subfact for a given fact by joining the subquery table.

## Checklist


docs/modules/ROOT/pages/references/architecture/metering-data-flow.adoc
```diff
@@ -257,6 +258,10 @@ For example, the ingestion job may need to sum all records belonging to a single
 | appuio_cloud_memory
 | appuio_cloud_memory_requested
 
+| is_subquery
```
corvus-ch, line 261: Now we reach the stage, where the naked SQL API might no longer be sufficient. If not taken care, values in this column and the content in the mapping table might drift apart. So I would prefer to not add it in the first place.

I assume, the reason for adding it is, it makes querying a lot easier (fewer joins). Are there any others?

corvus-ch (approved):
I would prefer to not introduce the `is_usbquery` field, but I understand why it might be added.

One of the upcoming features will be, to expose this data to the APPUiO Cloud Portal so that customer can access it. That would probably the time, where we hide this database behind an HTTP API. That would then mitigate my concern about the potential drift of data consistency.

glrf:
> One of the upcoming features will be, to expose this data to the APPUiO Cloud Portal so that customer can access it. That would probably the time, where we hide this database behind an HTTP API. That would then mitigate my concern about the potential drift of data consistency.

I would very much support introducing some kind of actual API to interface with the database (If you remember I wanted to do this from the start :sweat_smile: ). The queries are starting to get quite complex.

PR #137, opened 2022-11-17
add mysql to appcat exoscale dbaas documentation

susana-garcia:
## Summary

Add new mysql class to appcat exoscale dbaas documentation

## Checklist

- [X] Link this PR to related issues if applicable.

### DIFF
--- a/docs/modules/ROOT/pages/how-to/vshn-example/activate-exoscale-dbaas-appcat.adoc
+++ b/docs/modules/ROOT/pages/how-to/vshn-example/activate-exoscale-dbaas-appcat.adoc
@@ -35,12 +35,13 @@ WARNING: In case other AppCat services are already enabled on the {zone} such as
 parent="clusters/kv/${TENANT_ID}/${CLUSTER_ID}/appcat/exoscale/provider-exoscale"
 vault kv put "${parent}" access-key=<the-exoscale-access-key> secret-key=<the-exoscale-secret-key>
 ----
-. Configure cluster catalog
+. Configure cluster catalog and choose at least one of the following exoscale DBaaS:
 +
 [source,yaml]
 ----
 classes:
   - global.apps.appcat.dbaas.exoscale.postgres
+  - global.apps.appcat.dbaas.exoscale.mysql
 ----
 +
 This will deploy Crossplane and the Crossplane Exoscale provider.


ZuGao (approved):
LGTM

Kidswiss:
LGTM

Maybe one thought I had: would it make sense to introduce a class that handles all of dbaas for exoscale? So people don't have to import multiple classes every time?

Kidswiss (approved):

susana-garcia:
> LGTM
> 
> Maybe one thought I had: would it make sense to introduce a class that handles all of dbaas for exoscale? So people don't have to import multiple classes every time?

I see what you mean, but I wouldn't that. The chances that customers need all of them at the once may not be high, plus it can create confusion of what dbaas are included. But I may be wrong, please let me know your thoughts about it.

Kidswiss:
> I see what you mean, but I wouldn't that. The chances that customers need all of them at the once may not be high, plus it can create confusion of what dbaas are included. But I may be wrong, please let me know your thoughts about it.

My use-case is for appuio cloud, where we want all the services deployed. I'd actually argue that even on customer clusters everything should get deployed, even if they only use one service. Makes managing things easier.

susana-garcia:
> > I see what you mean, but I wouldn't that. The chances that customers need all of them at the once may not be high, plus it can create confusion of what dbaas are included. But I may be wrong, please let me know your thoughts about it.
> 
> My use-case is for appuio cloud, where we want all the services deployed. I'd actually argue that even on customer clusters everything should get deployed, even if they only use one service. Makes managing things easier.

Oh, now I see what you meant. In that case I agree.
Should I then leave the PR open for now?

Kidswiss:
> Oh, now I see what you meant. In that case I agree.
> Should I then leave the PR open for now?



> > > I see what you mean, but I wouldn't that. The chances that customers need all of them at the once may not be high, plus it can create confusion of what dbaas are included. But I may be wrong, please let me know your thoughts about it.
> > 
> > 
> > My use-case is for appuio cloud, where we want all the services deployed. I'd actually argue that even on customer clusters everything should get deployed, even if they only use one service. Makes managing things easier.
> 
> Oh, now I see what you meant. In that case I agree. Should I then leave the PR open for now?

No it's fine, it's something we'd have to plan and discuss first anyway.

You can go ahead and merge.

susana-garcia:
> > Oh, now I see what you meant. In that case I agree.
> > Should I then leave the PR open for now?
> 
> > > > I see what you mean, but I wouldn't that. The chances that customers need all of them at the once may not be high, plus it can create confusion of what dbaas are included. But I may be wrong, please let me know your thoughts about it.
> > > 
> > > 
> > > My use-case is for appuio cloud, where we want all the services deployed. I'd actually argue that even on customer clusters everything should get deployed, even if they only use one service. Makes managing things easier.
> > 
> > 
> > Oh, now I see what you meant. In that case I agree. Should I then leave the PR open for now?
> 
> No it's fine, it's something we'd have to plan and discuss first anyway.
> 
> You can go ahead and merge.

Ok, no problem.

PR #140, opened 2022-12-12
Add removal of appuio cloud organisation

DebakelOrakel:
## Checklist

- [ ] Try to isolate changes into separate PRs (to build a better changelog).
- [ ] Link this PR to related issues if applicable.


docs/modules/ROOT/pages/how-to/day2ops/remove-organisation.adoc
```diff
@@ -0,0 +1,17 @@
+= Remove an organisation
+
+== Prerequisites
+
+* All namespaces on every APPUiO Cloud zone, belonging to that organisation, is removed.
```
bastjan, line 5:
```suggestion
* All namespaces belonging to that organisation, on every APPUiO Cloud zone, are removed.
```

bastjan (commented):
It might be helpful to add the kubectl command to check there are no more organization namespaces.

We use the US spelling of `organi*z*ation`.

Also needs to be linked in the nav.

docs/modules/ROOT/pages/how-to/day2ops/remove-organisation.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+
+== Remove organisation
+
+
+[source,bash]
+----
+org=<NAME_OF_THE_ORG>
+
+kubectl delete organization $org
```
corvus-ch, line 16:
```suggestion
kubectl delete organization <name>
```

I kind of hate that notation. Actually, I find it dangerous. There can be a risk of the variable set to a wrong value when mixing up terminals.

docs/modules/ROOT/pages/how-to/day2ops/remove-organisation.adoc
corvus-ch, line 5: Please move that to the actual procedure and explain how to check.

```
kubectl get ns -o name -l appuio.io/organization=<name>
```

corvus-ch (changes requested):

docs/modules/ROOT/pages/how-to/day2ops/remove-organisation.adoc
DebakelOrakel: I agree, but thats how the docu is built :D

tobru:
Nit: Should we write `Organization` throughout the document (uppercase) because it's a name?

Is there something we have to check with billing / invoicing before we can safely delete the organization? Will Keycloak entries be removed as well? Are there any other artifacts to clean up?

corvus-ch:
> Is there something we have to check with billing / invoicing before we can safely delete the organization?

No, the mapping in the billing database must remain. I need to open an issue to give that mapping a time range to prevent confusion on reused names.

> Will Keycloak entries be removed as well? Are there any other artifacts to clean up?

By my understanding, the operator should remove all teams and also the main group when the Organization is deleted.

DebakelOrakel:
> Nit: Should we write `Organization` throughout the document (uppercase) because it's a name?

I'm ambivalent on this; personally I feel not but it could very well be right. We should ask a native speaker :D

docs/modules/ROOT/pages/how-to/day2ops/remove-organization.adoc
```diff
@@ -0,0 +1,26 @@
...
+* https://kb.vshn.ch/appuio-cloud/how-to/day2ops/connect-control-api.html[Connect to control api]
+
+
+== Check for organization namespaces
+
+On every APPUiO Cloud zone:
+
+[source,bash]
+----
+kubectl get ns -ocustom-columns="NAME:.metadata.name,ORG:.metadata.labels.appuio\.io\/organization" | grep <ORG_NAME>
```
corvus-ch, line 15:
```suggestion
kubectl get ns -o name -l appuio.io/organization=<ORG_NAME>
```

There is a simpler and more efficient way to do this. The `-l` option allows you to filter by label (server side, no need for grep or anything like that). With the above, only namespaces of that organization will be printed.

On a side node, the `-L` option will display a label as an additional column in the output. No need for the convoluted `-ocustom-columns`.

docs/modules/ROOT/pages/how-to/day2ops/remove-organization.adoc
```diff
@@ -0,0 +1,26 @@
...
+----
```
corvus-ch, line 16:
```suggestion
----

If any namespace is still present, contact the customer and make them delete it.
Alternatively, we can delete them but we need the agreement by the customer that it is OK.
Make them understand, that this will permanently delete all persistent volumes without a means to recover them.
```

I wonder what happens with K8up backups.

corvus-ch (changes requested):

corvus-ch:
> Should we write `Organization` throughout the document (uppercase) because it's a name?

We probably would need to differentiate between cases where we refer to the APPUiO Cloud specific construct or the Resource (capitalized) and where we refer to the concept in general (in the sense of the general language, small case). I usually lack the attention for such detail, though.

docs/modules/ROOT/pages/how-to/day2ops/remove-organization.adoc
```diff
@@ -0,0 +1,30 @@
+= Remove an organization
+
+== Prerequisites
+
+* All namespaces belonging to that organisation, on every APPUiO Cloud zone, are removed.
```
tobru, line 5:
```suggestion
* All namespaces belonging to that organization, on every APPUiO Cloud zone, are removed.
```

DebakelOrakel:
> > Should we write `Organization` throughout the document (uppercase) because it's a name?
> 
> We probably would need to differentiate between cases where we refer to the APPUiO Cloud specific construct or the Resource (capitalized) and where we refer to the concept in general (in the sense of the general language, small case). I usually lack the attention for such detail, though.

Spoke with Dawn; we should not use uppercase.

corvus-ch (approved):

PR #154, opened 2023-05-16
Add architecture documentation for usage profiles

simu:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/control-api-usage-profile.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/control-api-usage-profile.adoc
@@ -1,0 +1,114 @@
+= APPUiO Control API: UsageProfile
+
+TIP: This resource implements the xref:appuio-cloud:ROOT:references/quality-requirements/performance/resource-quota.adoc[] and xref:appuio-cloud:ROOT:references/quality-requirements/performance/ns-quota.adoc[] requirements.
+
+The `UsageProfile` encapsulates a set of quotas and policies which are applied to a namespace.
+Every organization has a `UsageProfile` assigned.
+When a new organization is created, a default `UsageProfile` is assigned.
+Depending on whether the new organization's `BillingEntity` has been validated, a different default `UsageProfile` is assigned.
+
+
+== Object
+
+.CRD based
+[source,yaml]
+----
+apiVersion: appuio.io/v1
+kind: UsageProfile
+metadata:
+  name: restricted <1>
+spec:
+  namespaceCount: 50 <2>
+  resources: <3>
+    organization-compute: <4>
+      apiVersion: v1
+      kind: ResourceQuota
+      spec:
+        hard:
+          limits.cpu: "8"
+          limits.memory: 20Gi
+          pods: "45"
+          requests.cpu: "4"
+          requests.memory: 4Gi
+        scopes:
+          - NotTerminating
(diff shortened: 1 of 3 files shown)


corvus-ch (commented):
My initial idea was, that usage profiles are just a piece of metadata. The rules would then be implemented in the zones. Having the usage profiles contain the relevant resources, is probably the simplest approach to implement. I like it.

Some things I am missing right now:

* One limitation we set, is the number of namespaces. It is nor yet clear to me if and how UsageProfiles can be used for this case.
* I would want to hide UsageProfiles from end users as much as possible, and UsageProfiles can only be changed by us. Field level permissions can be implemented with an admission web-hook. How does that impact API clients, especially the Cloud Portal? Would `kubectl edit` still work or would that require to use patch?
* Can we set usage profiles on BillingEntities which then gets cascaded to Organizations? What about having a default set on the BillingEntity, but a special override for just one specific Organization?

simu:
> My initial idea was, that usage profiles are just a piece of metadata. The rules would then be implemented in the zones. Having the usage profiles contain the relevant resources, is probably the simplest approach to implement. I like it.
> 
> Some things I am missing right now:
> 
> * One limitation we set, is the number of namespaces. It is nor yet clear to me if and how UsageProfiles can be used for this case.

Updated the PR to cover number of namespaces. This is fairly straightforward: define namespace count in usage profile & update the agent to reject namespace creation for an organization once the limit is hit.
 
> 
> * I would want to hide UsageProfiles from end users as much as possible, and UsageProfiles can only be changed by us. Field level permissions can be implemented with an admission web-hook. How does that impact API clients, especially the Cloud Portal? Would `kubectl edit` still work or would that require to use patch?

I've updated the documentation with an approach that should work without having to grant organization admins view on the usage profiles that their organiziations can use.

> 
> * Can we set usage profiles on BillingEntities which then gets cascaded to Organizations? What about having a default set on the BillingEntity, but a special override for just one specific Organization?

We can do that, at the expense of somewhat more complex logic in the controller. This could look roughly as follows:

* Add `spec.usageProfileRef` on the billing entity (store the value in Odoo the same way we currently store the BE `status` subresource) -- we'll need an admission webhook for BE updates which ensures BE admins can't edit the usage profile.
* Sync changes to a BE's `spec.usageProfileRef` to all organizations whose `spec.billingEntityRef` is the BE -- store in `status.usageProfileRef`
* Make the Organization's `spec.usageProfileRef`, optional, but give it preference over the BE's `spec.usageProfile` when setting the Organization's `status.usageProfileRef`.

Because we can't watch BEs for changes, we'd have to manually implement reconciliation logic (probably cronjob-style) to ensure that each organization's `status.usageProfileRef` is updated correctly whenever the associated BE's `spec.usageProfileRef` is updated. In the other direction, we'd also have to ensure an Organization's `status.usageProfileRef` is updated whenever the organization's `spec.billingEntityRef` is changed.

Note that the current proposal doesn't really cover how an organization's initial usage profile is defined. We discussed various options yesterday, one of them is that we'd define the usage profile at the BE level and propagate it to the organizations. Another option is to simply mark organizations as validated/verified or not, and select a default usage profile based on that flag (note that this has the same issue of needing to propagate the change from BE to organization).

bastjan (approved):
LGTM, i like the design!

corvus-ch (approved):

PR #157, opened 2023-08-25
Switch references from Thanos to Mimir

bastjan:
Also removes a section about setting up the receiver which is not required anymore.

## Checklist


haasad (approved):
Would it make sense to add a reference or link to the mimir equivalent of the section that was deleted? Or is the mimir setup not documented in this repo?

bastjan:
> Would it make sense to add a reference or link to the mimir equivalent of the section that was deleted? Or is the mimir setup not documented in this repo?

There is no need to do anything on the zone anymore. All setup is coming from the app.

haasad:
Perfect, thanks for the explanation

PR #159, opened 2023-10-18
Document system for providing metered billing data to Odoo16

HappyTetrahedron:
## Checklist


simu (approved):
LGTM overall, but it probably makes sense if at least a 2nd engineer and @corvus-ch review this as well.

tobru (approved):

docs/modules/ROOT/pages/references/architecture/metering-data-flow.adoc
```diff
@@ -1,730 +1,208 @@
...
+----
 
-| unit
-| MiB
-| MiB
-|===
+<1> Odoo ID of the Product or Product Variant
+<2> Identifier of this particular instance of the product
+<3> (Optional) Description as shown on invoice
+<4> Used for grouping lines item on the final invoice
```
corvus-ch, line 104:
```suggestion
<4> Used for grouping line items on the final invoice
```

docs/modules/ROOT/pages/references/architecture/metering-data-flow-appuio-cloud.adoc
```diff
@@ -0,0 +1,139 @@
...
+=== Extract
+
+From the static configuration, a number of query jobs are generated for each query rule and each billing time interval (default 1h).
+
+For a given rule and time interval, a separate job is created for each listed product.
+The product params are substituted into the query pattern, resulting in an unique query for each product, which only returns results for that specific product.
+
+The query must be formulated in such a way that the resulting timeseries contains the `sales_order_id` label, which determines the particular sales order to which this usage record should be billed.
+
+Each query result represents an usage data record, to be transformed and loaded into the target system.
```
corvus-ch, line 86:
```suggestion
Each query result represents a usage data record, to be transformed and loaded into the target system.
```

corvus-ch (changes requested):
Nice work as far as it concerns the content. From a structure perspective, we should put things in different locations.

* The generic description of the ETL data flow probably should go into https://docs.central.vshn.ch (Metered Billing > Data Flow).
* The page on APPUiO Managed, should go into https://docs.central.vshn.ch.

You built the content for APPUiO Managed on the one of APPUiO Cloud. But the product APPUiO Cloud is builds on top of APPUiO Managed. So the documentation for billing should also work in that order (for historical reasons, it was not the case so far).

This then also raises the question if `appuio_cloud_reporting` is the appropriate key. Maybe, `appuio_managed_reporting` is more suitable. Depending on how generic our implementation becomes, it might also be used for AppCat, so an even more generic name might be in order.

PR #164, opened 2024-02-29
Add decision page for scheduling billing jobs

HappyTetrahedron:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
+++ b/docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
@@ -25,7 +25,7 @@ To ensure continuous and correct reporting of resource usage, this ETL process n
 
 This describes the current implementation, as a baseline for comparison:
 
-A Project Syn Component produces a kube cronjob for each product that requires processing.
+A Project Syn component produces a Kubernetes cronjob for each product that requires processing.
 The cronjobs run hourly.
 The generated jobs run the ETL process for a relative timeframe ("3 hours ago until now").
 Failed jobs are exposed to monitoring via the KubeJobFailed metric.
@@ -49,7 +49,7 @@ if a single job run fails, but the next one succeeds, the successful job will ha
 
 === Option 2: Spawner Cronjob
 
-A Project Syn Component produces a single cronjob and provides it with a list of products for which the ETL process needs to run (including associated parameters).
+A Project Syn component produces a single cronjob and provides it with a list of products for which the ETL process needs to run (including associated parameters).
 The cronjob runs hourly.
 On each run, the cronjob generates a new job for each product which in turn runs the ETL process.
 These generated jobs run the ETL process for an absolute timeframe of 3 hours.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
```diff
@@ -0,0 +1,133 @@
...
+* Optimizing the ETL process itself
+* Detecting missing usage data directly in the target ERP system
+
+== Proposals
+
+=== Option 1: Cronjobs per Product
+
+This describes the current implementation, as a baseline for comparison:
+
+A Project Syn Component produces a kube cronjob for each product that requires processing.
```
simu, line 28:
```suggestion
A Project Syn component produces a Kubernetes cronjob for each product that requires processing.
```

docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
```diff
@@ -0,0 +1,133 @@
...
+.Con
+
+* Re-running a failed job does not produce expected results - it will process billing data for the current time, and not the time at which the job was supposed to run.
+* Even though the system can recover from transient errors, they still generate an alert in any case, and an investigation is required to determine whether manual intervention is necessary.
+* It is not easily possible to run the ETL process for an arbitrary timeframe.
+* Running on the order of 100 jobs each hour is taxing our infrastructure.
+
+=== Option 2: Spawner Cronjob
+
+A Project Syn Component produces a single cronjob and provides it with a list of products for which the ETL process needs to run (including associated parameters).
```
simu, line 52:
```suggestion
A Project Syn component produces a single cronjob and provides it with a list of products for which the ETL process needs to run (including associated parameters).
```

docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
```diff
@@ -0,0 +1,133 @@
...
+The cronjob runs hourly.
+On each run, the cronjob generates a new job for each product which in turn runs the ETL process.
```
simu, line 54: Would it be feasible to generate the jobs in such a fashion that we can circumvent [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) issues? For example by injecting configurable randomized delays (on the order of a few minutes) to smear the actual work of the jobs over 15 minutes or so.

This doesn't necessarily need to be outlined here, but we may want to consider this for the implementation.

simu (approved):
LGTM overall. Inline comments for some formatting/spelling nitpicks, and a random implementation idea that's probably too detailed for the decision page.

docs/modules/ROOT/pages/explanation/decisions/billing-scheduling-monitoring.adoc
HappyTetrahedron: Absolutely, and the implementation should do that.
To a degree, the issue is tamped by the job concurrency limit on the namespace. The pods run 5 at a time (even in the current implementation, which already does spawn them all at once)

